PR #300, opened 2021-07-19
Video Kit implementation + JPEG decoder update

Madour:
Video Kit implementation + JPEG decoder update from suggestions in #292


Madour:
Latest features updates:
- VideoKit can draw rectangles, texts and images
- Delete the useless old graphics classes (CGPixel, CGColor ...)
- Create graphic objects in gfx namespace : `gfx::Rectangle`, `gfx::Text`, `gfx::Image` and `gfx::Video`
- Updated lk_lcd spike, showcasing a video player with progress bar and time display

Internal updates:
- Make `CoreJPEGMode` a dependency of `CoreJPEG`
- Moved `CoreSDRAM` declaration outside of the VideoKit. In fact, SDRAM is not tied to VideoKit or CoreVideo and should be initialized at application level (ideally, should move it completely outside of CoreVideo driver in its own CoreMemory driver for example).
- Fix some bugs in DMA2D transfer and DSI refresh
- Can now choose DSI mode by setting `dsi::refresh_columns_count` value in `corevideo_config.h`:
    - 1 : default mode, no tearing effect callback, single column refresh
    - 2 : enable tearing effect report, and execute multiple partial refreshes by column
- Create `ComponentHandler<T>` to harmonize components declaration. All compoennts who have a `XXXHandler_TypeDef` have the same interface, and should inherit from `ComponentHandler<XXXHandlerTypeDef>` (replace XXX by DMA2D, JPEG, DSI ...)

PR #318, opened 2021-08-13
(spikes): imu stabilization with motors

SamHadjes:
POC for robot stabilization with IMU and motors

Only spike main.cpp file is really useful, all other additions are there to suppoort IMU driver which should be reimplemented


YannLocatelli:
Please refer to #444.

PR #401, opened 2022-01-05
✨ StateMachine with std::variant and eventFlags

YannLocatelli:


YannLocatelli:
All different tools used does not concluded quickly to a satisfying solution. Aborting this PR.

PR #430, opened 2022-01-19
ladislas/feature/logkit internal

ladislas:
- :recycle: (libs): LogKit - move implementation details into internal namespace
- :fire: (libs): LogKit - remove redundant leka::logger namespace


YannLocatelli (approved):
Rien à dire, tout roule

PR #420, opened 2022-01-18
hugo/features/rename LKCoreLL

ladislas:
- :truck: Rename LKCoreLL
- :truck: Remove prefix to LKCoreLightSensor
- :truck: Remove prefix LK to LKCoreMicrophone
- :truck: Remove prefix LK to LKCoreSTM32Hal
- :truck: Remove LK prefix to LKCoreFatFs
- :truck: Remove prefix LK to LKCoreI2CMock
- :truck: Remove prefix LK to LKCoreLuminositySensorTest
- :truck: Remove prefix LK to LKCoreOTM8009ATest
- :truck: Update prefix LK to LK Core comments

PR #432, opened 2022-01-19
:rotating_light: (tidy): Default init FIL _file object

ladislas:
Default init `FIL _file {}` to fix clang-tidy warnings.

I've made a separate PR instead of making the changes in #420 as I wanted to make sure it doesn't break anything.

It should not because the compiler would default initialize anyway but it's better to make it explicit.

### DIFF
--- a/drivers/CoreFs/include/CoreFatFs.h
+++ b/drivers/CoreFs/include/CoreFatFs.h
@@ -25,7 +25,7 @@ class CoreFatFs : public CoreFatFsBase
 	auto getPointer() -> FIL * final;
 
   private:
-	FIL _file;	 // TODO (@yann) - _file is not initialized - make it a pointer?
+	FIL _file {};
 };
 
 }	// namespace leka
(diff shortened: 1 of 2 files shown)


drivers/CoreFs/include/LKCoreFatFs.h
```diff
@@ -0,0 +1,32 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+#ifndef _LEKA_OS_LIB_FATFS_H_
+#define _LEKA_OS_LIB_FATFS_H_
+
+#include "LKCoreFatFsBase.h"
+
+namespace leka {
+
+class LKCoreFatFs : public LKCoreFatFsBase
```
YannLocatelli, line 12: As reminder

```suggestion
class CoreFatFs : public CoreFatFsBase
```

Ainsi que le nom du fichier

YannLocatelli (changes requested):

YannLocatelli (approved):

PR #460, opened 2022-01-25
✨ (timeout): Add CoreTicker and CoreTimeout

YannLocatelli:
CoreTimeout will help to switch into Sleep state in the StateMachine


ladislas (approved):
super clean! trop cool 👍

PR #561, opened 2022-03-13
✨ RobotController - Add ble service device information

YannLocatelli:


ladislas (approved):
LGTM 👍 

juste une question : le `inline static` est obligatoire?

PR #597, opened 2022-03-17
ladislas/feature/sm add logger

ladislas:
- :loud_sound: (sm): Add logger to rc's state machine
- :lipstick: (rc): Clean up sm logs from namespace garbage


libs/RobotKit/include/RobotController.h
```diff
@@ -222,7 +222,9 @@ class RobotController : public interface::RobotController
 	}
 
   private:
-	std::chrono::seconds _sleep_timeout_duration {300};
+	system::robot::sm::logger logger {};
+
+	std::chrono::seconds _sleep_timeout_duration {10};
```
YannLocatelli, line 227: Previously set at 300 seconds before enter sleeping state, seems not relevant to modify it with this PR

```suggestion
	std::chrono::seconds _sleep_timeout_duration {300};
```

libs/RobotKit/include/RCLogger.h
```diff
@@ -0,0 +1,72 @@
...
+		auto name = trim_namespaces(boost::sml::aux::get_type_name<T>());
+		return name;
+		// return boost::sml::aux::get_type_name<T>();
+	}
+
+	template <typename SM, typename TEvent>
+	void log_process_event([[maybe_unused]] const TEvent &event)
+	{
+		auto e = name<TEvent>();
+		log_info("%s", e.c_str());
```
YannLocatelli, line 45: Why use `log_info` when previously used `log_debug`?

YannLocatelli (approved):
Except question/detail, LGTM

libs/RobotKit/include/RobotController.h
ladislas: yup! sorry about that, corrected!

libs/RobotKit/include/RCLogger.h
ladislas: because log_debug is only for debug, thus it must be removed before being merged.

log_info can be used in production to "know" what's happening. it's only for the important things and must be seldom used.

if/when we use log files, only info, errors and warnings (if added) will be saved

PR #602, opened 2022-03-17
🚸 (video): Add turn on/off + Play in a loop

YannLocatelli:


libs/VideoKit/include/VideoKit.h
```diff
@@ -13,9 +13,12 @@ class VideoKit
 
 	void initializeScreen() {};
 
+	void turnOn() {};
+	void turnOff() {};
+
 	void displayImage(const char *path) {};
 
-	void playVideo(const char *path) {};
+	void playVideo(const char *path, bool play_in_a_loop = false) {};
```
YannLocatelli, line 21:
```suggestion
	void playVideo(const char *path, bool must_loop = false) {};
```

PR #629, opened 2022-03-24
✅ (EventQueue): Use equeue_stub to call callbacks

YannLocatelli:


libs/BatteryKit/source/BatteryKit.cpp
```diff
@@ -19,7 +19,6 @@ void BatteryKit::startEventHandler()
 		}
 	};
 
-	on_tick();	 // TODO (@john_doe): only for unit tests. Due to event_queue that does not make the call
 	_event_queue.call_every(1s, on_tick);
```
YannLocatelli, line 22: On peut se poser la question de garder `CoreEventQueue`

PR #633, opened 2022-03-25
⚡️ (ble): Set dynamic data in advertising (name, battery level, battery is charging)

YannLocatelli:


libs/BLEKit/source/CoreGap.cpp
```diff
@@ -5,14 +5,17 @@
 using namespace leka;
 using namespace ble;
 
+void CoreGap::setAdvertisingData(CustomAdvertising data)
+{
+	_advertising_data_builder.setName(data.name);
+	_advertising_data_builder.setServiceData(
```
YannLocatelli, line 11: Note: Only one ServiceData is allowed, one uuid with an array of data

libs/BLEKit/source/CoreGap.cpp
```diff
@@ -5,14 +5,17 @@
...
+		leka::service::commands::uuid, {{data.battery, data.is_charging, data.led_red, data.led_green, data.led_blue}});
```
YannLocatelli, line 12: Note: Only one ServiceData is allowed, one uuid with an array of data

libs/BLEKit/include/BLEKit.h
```diff
@@ -22,12 +22,19 @@ class BLEKit
 
 	void init();
 
+	void setName(const char *name);
+	void setBatteryLevel(uint8_t value);
+	void setIsCharging(bool value);
+	void setDisplayedColor(uint8_t red_channel, uint8_t green_channel, uint8_t blue_channel);
+
```
ladislas, line 29: Having BLEKit know about name, battery, charging state and displayed color breaks the SRP.

To fix that, you have two methods:

- `void setAdvertisingData(AdvertisingData & data)`
- `auto getAdvertisingData() -> AdvertisingData`

This way, the `AdvertisingData` struct can evolve without impacting BLEKit

libs/BLEKit/include/CoreGap.h
```diff
@@ -13,6 +13,15 @@
...
 namespace leka {
 
+struct CustomAdvertising {
+	const char *name;
+	uint8_t battery;
+	uint8_t is_charging;
+	uint8_t led_red;
+	uint8_t led_green;
+	uint8_t led_blue;
+};
```
ladislas, line 23: to follow on the previous comment about SRP, you could do something like this:

```suggestion
struct AdvertisingData {
	const char *name;
	uint8_t battery;
	uint8_t is_charging;
	RGB color;
    
	auto getData() -> std::span <uint8_t> {
		auto tmp  = std::to_array<uint8_t> ({battery, is_charging, color.red, color.green, color.blue});
		std::copy(tmp.begin(), tmp.end(), data.begin());
		return std::span<uint8_t> {data.data(), std::size(tmp)};
	}
		
	private:
		std::array<uint8_t, 16> data {};
};
```

see https://gcc.godbolt.org/z/GK7P6KhG6

ladislas: this encapsulate the advertising data logic inside the struct making it easy to change with minimum (if any) impact on the code using it.

libs/BLEKit/include/CoreGap.h
```diff
@@ -24,11 +33,12 @@ class CoreGap
...
 	// void onInit(std::function<void()> cb) { _post_init = cb; }
 
-	void setDeviceName(const char *name);
-
 	void startAdvertising();
+	void refreshAdvertising(CustomAdvertising custom_advertising);
 
   private:
+	void setAdvertisingData(CustomAdvertising data);
+
```
ladislas, line 41: `refresh` is a bit strange. it gives the false impression that it's going to pull the new data on it's own, like when you refresh a page in your browser.

I would have only one method, `setAdvertisingData`. 

if they change, we just set them again.

ladislas (commented):
First read through, it's very interesting 👍

libs/BLEKit/include/CoreGap.h
YannLocatelli: `getData` will replace a part of process in `setAdvertisingData` in `CoreGap` and its indeed a good way to set data without knowing them in CoreGap

Since there is a function in the struct, I will maybe move it in another file + add unit tests to check its function

Also note that due addition of `std::array data`, it is no more possible to initialize a struct using `AdvertisingData{.name = "Leka"};`

libs/BLEKit/include/CoreGap.h
YannLocatelli: `refresh` refers to refresh dynamically the data in advertising.

The `setAdvertisingData` used in CoreGap will set in local only the data from the struct into `advertising_data_builder` from mbed

`refresh` will do this plus a call to `_gap.setAdvertisingPayload(_advertising_handle, _advertising_data_builder.getAdvertisingData());` to set data from local to "online"

libs/BLEKit/include/BLEKit.h
YannLocatelli: Yes, but not sure which gonna store it, BLEKit or Coregap. In the last case, BLEKit will be only a passthrough.

ladislas: it cas be in CoreGap then

libs/BLEKit/include/CoreGap.h
ladislas: this has been discussed directly on slack, are we good on the issue?

libs/BLEKit/include/CoreGap.h
ladislas: Maybe I'm missing something, but I don't see a case where you want to refresh the offline data but not make them available to the outside world.

I would refactor CoreGap to have setAdvertisingData do the offline + online work, if that makes sens.

This can be tackled in an upcoming PR of course

YannLocatelli: If you just want to replace the name `refresh` by `setAdvertisingData`, no problem

libs/BLEKit/include/AdvertisingData.h
```diff
@@ -0,0 +1,34 @@
...
+	std::array<uint8_t, 16> data {};
+
+	auto getData() -> std::span<const uint8_t>
+	{
+		auto tmp = std::to_array<uint8_t>({battery, is_charging, led_red, led_green, led_blue});
+		std::copy(tmp.begin(), tmp.end(), data.begin());
+		return std::span<uint8_t> {data.data(), std::size(tmp)};
+	}
+
+	[[nodiscard]] auto getDataSize() const -> int { return _advertising_data_size; }
```
ladislas, line 31: I was not happy with my previous solution. Your solutions in particular showed me the shortcomings of mine.
so I spent a little more time trying to find a way that would make it more future proof, avoid issues and work in our uses cases.

I've come up with the following, the only difference being the colors that you can keep the way you made them as discussed.

```cpp
 struct AdvertisingData {
    AdvertisingData(const char* n, uint8_t b, uint8_t ic, RGB c) : name(n), battery(b), is_charging(ic), color(c) {}

	const char *name {};
	uint8_t battery {};
	uint8_t is_charging {};
	RGB color {};

    auto values() {
        updateValues();
        return std::span{_values};
    }
    
    auto data() {
        updateValues();
        return std::span{_values}.data();
    }

    auto size() { return _values.size(); }

    private:
    void updateValues() {
         _values = {battery, is_charging, color.red, color.green, color.blue};
    }

    std::array<uint8_t, 5> _values = {};
};
```

the goal was:

- avoid the need to store the size
- have a well known interface with `data()` and `size()`
- have `data()` actually do what `std::array::data()` does --> pointer to first element
- have `values()` that returns a span to be able to iterate over the values
- hide the implementation details and underlying container
- be able to update one value and still get all the values with the updated one

You can see a working example with comments here: https://gcc.godbolt.org/z/WssxGv7qs

libs/BLEKit/include/CoreGap.h
```diff
@@ -24,9 +26,8 @@ class CoreGap
 	void onInitializationComplete(BLE::InitializationCompleteCallbackContext *params);
 	// void onInit(std::function<void()> cb) { _post_init = cb; }
 
-	void setDeviceName(const char *name);
-
 	void startAdvertising();
+	void refreshAdvertising(AdvertisingData advertising_data);
```
ladislas, line 30: can we have

```suggestion
	void refreshAdvertising(const AdvertisingData &advertising_data);
```
?

libs/BLEKit/source/CoreGap.cpp
```diff
@@ -7,12 +7,12 @@ using namespace ble;
...
-	_advertising_data_builder.setManufacturerSpecificData({{0x2A, 0x2B, 0x2C, 0x2D}});
-	_advertising_data_builder.setAdvertisingInterval(adv_interval_t::min());
-	_advertising_data_builder.setServiceData(GattService::UUID_BATTERY_SERVICE, {{0x42}});
-	_advertising_data_builder.setServiceData(leka::service::commands::uuid, {{0}});
+	auto default_advertising_data = AdvertisingData {};
+
+	_advertising_data_builder.setName(default_advertising_data.name);
+	_advertising_data_builder.setServiceData(
+		leka::service::commands::uuid,	 // TODO: commands::uuid only for compatibility with LekaApp
+		{default_advertising_data.getData().data(), default_advertising_data.getDataSize()});
```
ladislas, line 15: with my new implementation, no need for `getData().data()`

```suggestion
		{default_advertising_data.data(), default_advertising_data.size()});
```

libs/BLEKit/source/CoreGap.cpp
```diff
@@ -7,12 +7,12 @@ using namespace ble;
 
 void CoreGap::setDefaultAdvertising()
```
ladislas, line 8: question: instead of having two methods which do almost the same thing, `setDefaultAdvertising` the first time and then `refreshAdvertising`, why not just `refreshAdvertising(ble::advertising::default_data)`?

ladislas (commented):
new review. a few suggestions/questions, and we're good to merge

libs/BLEKit/include/CoreGap.h
YannLocatelli: This error appears

``` bash
/Users/yann/dev/LekaOS/libs/BLEKit/source/CoreGap.cpp:47:26: error: passing 'const leka::AdvertisingData' as 'this' argument discards qualifiers [-fpermissive]
   47 |   {advertising_data.data(), advertising_data.size()});
```

libs/BLEKit/source/CoreGap.cpp
YannLocatelli: It compiles but clangd shows this error

```
Non-constant-expression cannot be narrowed from type 'std::array::size_type' (aka 'unsigned long') to 'mbed::Span<type-parameter-0-0, -1>::index_type' (aka 'long') in initializer list (fix available)clang(-Wc++11-narrowing)
```

PR #725, opened 2022-04-26
hugo/feature/Add RFID state machine

HPezz:
- :truck: (drivers): Rename IsTagDetected
- :sparkles: (libs): Add RFIDStateMachine & make public some funcs


HPezz:
Deprecated version of RFID SM, refer to new version in #929

PR #801, opened 2022-05-20
⚡️ (video): Do not display image if already displayed

YannLocatelli:
- [x] Validated on robot (20/05/2022)

---

Note about validation: In a spike, display same image n-times in a row and log in displayImage function something to assert program reach this line only one time


ladislas (approved):
LGTM 👍

PR #848, opened 2022-06-07
:adhesive_bandage: (sm): Move actions in "on_entry"

HPezz:


ladislas:
not needed anymore, maybe in the future.

PR #849, opened 2022-06-07
🚸 (rfid): Turn off screen on emergency stop

YannLocatelli:


YannLocatelli:
Notice that screens of reinforcers are not turned on

PR #918, opened 2022-06-25
:art: (cmake): Reformat all CMake files

ladislas:
Recent update in vscoce's CMake Tools added automating formating

Instead of fighting the tools, we reformatted all files through the tool
to avoid future conflicts


ladislas:
replace by #919

PR #1003, opened 2022-09-07
:bug: (tools): Fix args namespace case

ladislas:


tools/check_fs_content.py
```diff
@@ -159,7 +159,7 @@ def print_report(broken, empty, missing, reboot):
 
 # Relative path from LekaOS root
 ROOT_PATH = args.dir if not args.test else 'fs/usr/test'
-LOOP_SLEEP_DELAY = args.SLEEP_DELAY if not args.FAST else 0.2
+LOOP_SLEEP_DELAY = args.sleep_delay if not args.fast else 0.2
```
HPezz, line 162:
```suggestion
LOOP_SLEEP_DELAY = args.loop_delay if not args.fast else 0.2
```

tools/check_fs_content.py
```diff
@@ -159,7 +159,7 @@ def print_report(broken, empty, missing, reboot):
 
 # Relative path from LekaOS root
 ROOT_PATH = args.dir if not args.test else 'fs/usr/test'
-LOOP_SLEEP_DELAY = args.SLEEP_DELAY if not args.FAST else 0.2
+LOOP_SLEEP_DELAY = args.loop_delay if not args.fast else 0.2
```
YannLocatelli, line 162: Just wondering why `metavar` cannot be used, but it's only used on help command so that is fine [[Link](https://docs.python.org/3/library/argparse.html#metavar)]

YannLocatelli (approved):
- [x] Validated on board on 22/09/2022

PR #1017, opened 2022-09-12
:recycle: (FirmwareKit): Make FirmwareKit independent of path, format

ladislas:
- add config to define path and format
- changed the ctor to accept config (use DI)
- define default config
- test case where os version is not available
- clean up tests with global variables


YannLocatelli (approved):
- [x] Validated on robot, with update

PR #1022, opened 2022-09-14
:recycle: (videokit): Add a display_white_background option in displayImage

HPezz:


libs/VideoKit/tests/VideoKit_test.cpp
```diff
@@ -51,6 +51,14 @@ TEST_F(VideoKitTest, displayImage)
 	video_kit.displayImage(temp_file_path);
 }
 
+TEST_F(VideoKitTest, displayImageWithBackground)
+{
+	EXPECT_CALL(mock_event_flags, set(VideoKit::flags::STOP_VIDEO_FLAG)).Times(2);
+	EXPECT_CALL(mock_corevideo, displayImage).Times(2);
```
YannLocatelli, line 57: Perfect, eventually enforce the call of `displayImage` only once for the test above

libs/VideoKit/source/VideoKit.cpp
```diff
@@ -29,14 +29,24 @@ void VideoKit::initializeScreen()
...
 
+	if (display_white_backgroung) {
+		_event_flags.set(flags::STOP_VIDEO_FLAG);
+		auto file = FileManagerKit::File {"fs/home/img/system/robot-misc-robot-misc-screen_empty_white.jpg"};
+
+		rtos::ThisThread::sleep_for(100ms);
+
+		_video.displayImage(file);
+		file.close();
+	}
```
YannLocatelli, line 48: Must arrived after `_event_flags.set(flags::STOP_VIDEO_FLAG);`, otherwise the video will play, flash and stop with something different of a _empty white screen_.

Also let a delay before display it, 'cause of time for video thread to stop after the raise of STOP_VIDEO_FLAG

include/interface/libs/VideoKit.h
```diff
@@ -15,7 +15,7 @@ class VideoKit
 
 	virtual void initializeScreen() = 0;
 
-	virtual void displayImage(const std::filesystem::path &path) = 0;
+	virtual void displayImage(const std::filesystem::path &path, bool display_white_backgroung = false) = 0;
```
YannLocatelli, line 18: My suggestion of name: 

```suggestion
	virtual void displayImage(const std::filesystem::path &path, bool fill_background = false) = 0;
```
```suggestion
	virtual void displayImage(const std::filesystem::path &path, bool fill_with_white_background = false) = 0;
```

ladislas (approved):

PR #1023, opened 2022-09-14
🔨 (version): Move firmware version from build_firmware script to Makefile

YannLocatelli:
- [x] Validation on `application-signed` file + robot with github CI on 20/09/2022

Add regex to check format of version

Suggest that build come in another PR


tools/firmware/build_firmware.sh
```diff
@@ -5,22 +5,29 @@ mkdir -p _release
...
 APPLICATION_SIGNED_HEX="_tmp/application-signed.hex"
 
 FIRMWARE_HEX="_release/firmware.hex"
 FIRMWARE_RELEASE="_release/firmware.bin"
 
+APPLICATION_VERSION_REGEX="^[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,5}(\+[0-9]+)?$"
+if [ -z "$APPLICATION_VERSION" ] || [[ ! $APPLICATION_VERSION =~ $APPLICATION_VERSION_REGEX ]];
+then
+    echo "Invalid version, exiting script..."
+    exit 0
```
ladislas, line 28: because it's a failure, the script must return 1, not 0

in shells, 0 == success, anything else == failure

```suggestion
    exit 1
```

tools/firmware/build_firmware.sh
```diff
@@ -5,22 +5,29 @@ mkdir -p _release
...
 BOOTLOADER_HEX="_tmp/bootloader.hex"
 
 APPLICATION_HEX_SOURCE="_build/LEKA_V1_2_DEV/app/os/LekaOS.hex"
-APPLICATION_VERSION="1.1.0+0"
 APPLICATION_SIGNED_HEX="_tmp/application-signed.hex"
 
 FIRMWARE_HEX="_release/firmware.hex"
 FIRMWARE_RELEASE="_release/firmware.bin"
 
+APPLICATION_VERSION_REGEX="^[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,5}(\+[0-9]+)?$"
```
ladislas, line 24: nice! 

I think it would be even better to use the official semver regex to handle all possibles cases:

https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string

https://regex101.com/r/vkijKf/1/

```suggestion
// ? Official semver regex, see
// ? https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string
APPLICATION_VERSION_REGEX="^(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(?:-((?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+([0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$"

```

Makefile
```diff
@@ -24,6 +24,7 @@ MCUBOOT_VERSION ?= $(shell cat $(ROOT_DIR)/config/mcuboot_version)
 BAUDRATE        ?= 115200
 BUILD_TYPE      ?= Release
 TARGET_BOARD    ?= LEKA_V1_2_DEV
+FIRMWARE_VERSION?= 1.1.0
```
ladislas, line 27: using the make file to run the script is a good idea, but the makefile itself cannot define the version.

just like for mbed-os version, the firmware version must live in a separate file like  `config/leka_os_version`

then use the `FIRMWARE_VERSION?=$(shell cat $(ROOT_DIR)/config/leka_os_version)`

ladislas (changes requested):
Thanks for the PR, LGTM 👍 

I like the fact that we are using `make firmware` but the version itself should be defined in a config file, not in the Makefile

tools/firmware/build_firmware.sh
YannLocatelli: ok

tools/firmware/build_firmware.sh
YannLocatelli: But it is not compliant with mcuboot and might broke it.

Makefile
YannLocatelli: ok

tools/firmware/build_firmware.sh
ladislas: ah, in what way is it not compliant?

YannLocatelli: Here a sample of match for this regex:

> 1.0.0-alpha.1
1.0.0-alpha0.valid
1.0.0-alpha.0valid
1.0.0-alpha-a.b-c-somethinglong+build.1-aef.1-its-okay
1.0.0-rc.1+build.1
2.0.0-rc.1+build.123
1.2.3-beta
10.2.3-DEV-SNAPSHOT
1.2.3-SNAPSHOT-123
1.0.0
2.0.0
1.1.7
2.0.0+build.1848
2.0.1-alpha.1227
1.0.0-alpha+beta
1.2.3----RC-SNAPSHOT.12.9.1--.12+788
1.2.3----R-S.12.9.1--.12+meta
1.2.3----RC-SNAPSHOT.12.9.1--.12
1.0.0+0.build.1-rc.10000aaa-kk-0.1
99999999999999999999999.999999999999999999.99999999999999999
1.0.0-0A.is.legal

Some works as `1.0.0` or `2.0.0`, but I already know that `99999999999999999999999.999999999999999999.99999999999999999` will fail since mcuboot cannot handle values above 1 byte for major and minor and 2 bytes for revision (it happened)

For the remaining, I can try them all if you want, but I am pretty sure that at least one will not fit, maybe `1.0.0-alpha-a.b-c-somethinglong+build.1-aef.1-its-okay` or `1.0.0-0A.is.legal`

Makefile
```diff
@@ -15,16 +15,16 @@ MCUBOOT_DIR       := $(ROOT_DIR)/extern/mcuboot
...
+PORT             ?= /dev/tty.usbmodem14303
+MBED_GIT_URL     ?= $(shell cat $(ROOT_DIR)/config/mbed_git_url)
+MBED_BRANCH      ?= $(shell cat $(ROOT_DIR)/config/mbed_version)
+MBED_VERSION     ?= $(shell cat $(ROOT_DIR)/config/mbed_version)
+MCUBOOT_GIT_URL  ?= $(shell cat $(ROOT_DIR)/config/mcuboot_git_url)
+MCUBOOT_VERSION  ?= $(shell cat $(ROOT_DIR)/config/mcuboot_version)
+BAUDRATE         ?= 115200
+BUILD_TYPE       ?= Release
+TARGET_BOARD     ?= LEKA_V1_2_DEV
+FIRMWARE_VERSION ?= $(shell cat $(ROOT_DIR)/config/leka_os_version)
```
ladislas, line 27: nice! thanks :)

tools/firmware/build_firmware.sh
YannLocatelli: A script to check os version prior to give it to Makefile

YannLocatelli:
Suggest that build come in another PR

Makefile
```diff
@@ -15,15 +15,16 @@ MCUBOOT_DIR       := $(ROOT_DIR)/extern/mcuboot
...
```
YannLocatelli, line 27: Better to use `OS_VERSION` here

```suggestion
OS_VERSION ?= $(shell cat $(ROOT_DIR)/config/leka_os_version)
```

tools/firmware/build_firmware.sh
YannLocatelli: might by removed if script that check semver **and** mcuboot is accepted

Makefile
```diff
@@ -259,6 +264,12 @@ clang_tidy_diff:
 		| grep -E "\.h\$$|\.cpp\$$"            \
 		| xargs --no-run-if-empty ruby tools/run-clang-tidy.rb $(CMAKE_TOOLS_BUILD_DIR)
 
+check_os_version_valid:
+	@echo ""
+	@echo "🔨 Checking os version is valid ☑"
+	@echo ""
+	python3 tools/check_version.py $(FIRMWARE_VERSION) --standards semver mcuboot
```
ladislas, line 271: I'm not sure we need this check added to the Makefile. It is highly specific and not useful on a day to day basis.

I would prefer we call the script directly, like we do for `check fs content`

tools/check_version.py
```diff
@@ -0,0 +1,94 @@
...
+list_of_standards = {"semver": semver_standard,
+                     "mcuboot": mcuboot_standard}
+
+#
+# MARK: - argparse
+#
+
+parser = argparse.ArgumentParser(description='Check version is valid with standard')
+
+parser.add_argument('input_version', metavar='input_version', type=str,
```
ladislas, line 54: I would pass the file by default, instead of the version

this way we can use the script in CI to make sure that version bumps are compliant.

`./tools/check_os_version.py ./config/leka_os_version`

Makefile
ladislas: why is that more accurate?

config/leka_os_version
```diff
@@ -0,0 +1 @@
+1.1.0
```
ladislas, line 1: I'm wondering if `os_version` is not better, than `leka_os_version` especially if we change the name of the variable in the Makefile

tools/check_version.py
```diff
@@ -0,0 +1,94 @@
...
+        self.error_message:str = error_message
+
+# Source: https://semver.org/
+semver_regex = "^(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(?:-((?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+([0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$"
+semver_standard = Standard(is_version_compliant=lambda version: re.match(semver_regex, version) is not(None),
+                           success_message="✅ Version is compliant with Semantic Versioning 2.0.0.",
+                           error_message="❌ Version is not compliant with Semantic Versioning 2.0.0.")
+
+
+def mcuboot_is_version_compliant(version: str):
```
ladislas, line 26:
```suggestion
def version_is_mcuboot_compliant(version: str):
```

tools/check_version.py
```diff
@@ -0,0 +1,94 @@
...
+
+class Standard:
+    def __init__(self, is_version_compliant:Callable[[str], bool], success_message:str, error_message:str):
+        self.is_version_compliant:Callable[[str], bool] = is_version_compliant
+        self.success_message:str = success_message
+        self.error_message:str = error_message
+
+# Source: https://semver.org/
+semver_regex = "^(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(?:-((?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+([0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$"
+semver_standard = Standard(is_version_compliant=lambda version: re.match(semver_regex, version) is not(None),
```
ladislas, line 21: I would create a function `version_is_semver_compliant` here instead of using a lambda, for consistency with what comes next and to make it clearer to the reader

tools/check_version.py
```diff
@@ -0,0 +1,94 @@
...
+
+#
+# MARK: - Standards
+#
+
+class Standard:
+    def __init__(self, is_version_compliant:Callable[[str], bool], success_message:str, error_message:str):
+        self.is_version_compliant:Callable[[str], bool] = is_version_compliant
+        self.success_message:str = success_message
+        self.error_message:str = error_message
```
ladislas, line 17: why is this needed?

I'm not a python expert, but I find this very hard to read and I'm not sure I understand the purpose

tools/check_version.py
```diff
@@ -0,0 +1,94 @@
...
+                            error_message="❌ Version is not compliant with MCUboot.")
+
+list_of_standards = {"semver": semver_standard,
+                     "mcuboot": mcuboot_standard}
+
+#
+# MARK: - argparse
+#
+
+parser = argparse.ArgumentParser(description='Check version is valid with standard')
```
ladislas, line 52:
```suggestion
parser = argparse.ArgumentParser(description='Check os version compliance with standards used in the project')
```

ladislas (changes requested):
Thanks for the changes, this is much better 👍 

I find the python script a bit over engineered/complicated to be honest.

I'm not sure we need to pass arguments, especially the standards, and even the input could be hardcoded (see my comment about using a path instead of the string for CI)

The checks are quite simple, super fast, so scripting/functional programming should be preferred to OOP.

Let's discuss that on Monday :)

tools/check_version.py
```diff
@@ -0,0 +1,99 @@
...
+
+
+def getDataFromFile(file_path: str) -> str:
+    data: str = ""
+    try:
+        f = open(file_path, "r")
+        data = f.read()
+        f.close()
+    except BaseException as error:
+        print(f"{error}")
```
YannLocatelli, line 68:
```suggestion
        print(f"{error}")
        sys.exit(1)
```

YannLocatelli:
- [x] Validation on `application-signed` file on 20/09/2022

PR #1025, opened 2022-09-14
🔧 (version): FirmwareKit use os version generated from cmake

YannLocatelli:
Need: 

- [x] #1023
- [ ] Add workflow to check version are correspondant in config and generated binary
- [x] Rename `Firmware_Version` by `Version`
- [x] ~~Fix use of `settings.json`~~
- [ ] Refactor getCurrentVersionFromFile


CMakeLists.txt
```diff
@@ -65,6 +65,13 @@ if(ENABLE_LOG_DEBUG AND ENABLE_SYSTEM_STATS)
 	add_definitions(-DMBED_HEAP_STATS_ENABLED)
 endif(ENABLE_LOG_DEBUG AND ENABLE_SYSTEM_STATS)
 
+# Firmware version
+option(FIRMWARE_VERSION "Set firmware version" "0.0.0")
+
+if(FIRMWARE_VERSION)
```
ladislas, line 71: as you give the initial value `0.0.0` to `FIRMWARE_VERSION`, the condition is always true.
the `if` can be removed

libs/FirmwareKit/source/FirmwareKit.cpp
```diff
@@ -10,7 +10,12 @@ using namespace leka;
 
 auto FirmwareKit::getCurrentVersion() -> FirmwareVersion
 {
+#if defined(FIRMWARE_VERSION)
```
ladislas, line 13: I'm not a big fan of macros, especially this deep in the implementation. 

my understanding is that `#if defined(FIRMWARE_VERSION)` is always true, even if the value is the empty string `""`

If I'm not mistaken, then the other path is never reached and can be removed.

what is the rational for keeping the `getCurrentVersionFromFile` part?

I'm really careful with these macro magic tricks. My take would be "force the firmware version everywhere, there is no reason to not define/use it and skip all the checks".

ladislas (changes requested):
Looks nice but I'm worried about the use of macros like that. To be discussed tomorrow :)

CMakeLists.txt
YannLocatelli: ok

libs/FirmwareKit/source/FirmwareKit.cpp
YannLocatelli: > my understanding is that #if defined(FIRMWARE_VERSION) is always true, even if the value is the empty string ""

Not it is not, if it is the case I will not implement it this way.

But I get the point of the reliability of macro, so let me know what is your suggestion

YannLocatelli:
- [ ] Add workflow to check version are correspondant in config and generated binary
- [ ] Rename `Firmware_Version` by `Version`

YannLocatelli:
Replaced by #1069

PR #1026, opened 2022-09-15
🧱 (serialnumber): Handle epoch format for date_of_test

YannLocatelli:
- [x] Validated on robot on 20/09/2022 (old and new date_of_test format)

Need

- [x] #946 
- [x] Determine between `std::atoll` and `std::strftime` increase flash → is `std::strftime`
- [x] epoch in plain text or bytes → plain text


YannLocatelli:
Determine between `std::atoll` and `std::strftime` increase flash

YannLocatelli:
- [x] Validated on robot on 20/09/2022

ladislas (approved):

PR #1035, opened 2022-09-20
:recycle: (activity): Invert minus & plus display

HPezz:


ladislas (approved):

ladislas:
bug filed here: #1036

PR #1048, opened 2022-09-29
:recycle: (MagicCard): Add language byte to MagicCard

HPezz:


libs/RFIDKit/include/MagicCard.h
```diff
@@ -30,6 +30,8 @@ struct MagicCard {
 		return both;
 	}
 
+	[[nodiscard]] constexpr auto getLanguage() const -> uint8_t { return _tag.data.at(language_byte_index); }
```
ladislas, line 33: not asking for a change, but after re-reading the whole code, having at `getXXX` feels a bit verbose. It's just a simple struct, `card.id()` or `card.language()` would be good enough.

to keep in mind for a future refactor ;)

ladislas (approved):
LGTM 👍

libs/RFIDKit/include/MagicCard.h
YannLocatelli, line 33: Instead of returning a byte (0x01, 0x02, ...) I think it is better to wrap it into an Enum to understand what is happening on use

``` cpp
enum class TagLanguage{
	None = 00,
	fr_FR = 01,
	en_US = 02
}
```

---

```cpp
	if (card.getLanguage() == TagLanguage::fr_FR) {
		_videokit.displayImage("/fs/home/img/system/robot-misc-choose_activity-fr_FR.jpg");
	} else if (card.getLanguage() == TagLanguage::en_US){
		_videokit.displayImage("/fs/home/img/system/robot-misc-choose_activity-en_US.jpg");
	}
```

YannLocatelli: Going further, we can have a python dictionary like or struct as `Color.h` in ActivityKit to get language tag from the byte in order to directly set the right tag in the path

`_videokit.displayImage({image_system_folder} + {id_converted_name} + '-' + {tag_language} + ".jpg");`

We can abord this when we will simplify use of videokit by dropping the folder and extension, {image_system_folder} and ".jpg" in the above case.

ladislas: I agree with @YannLocatelli 

Just a small note: I would call the enum just `Language` and keep in `struct MagicCard` to access/use it via `MagicCard::Language`

HPezz: To make it more readable and using Language inside MagicCard's architecture, use `uint16_t _id` & `Language _language` as variables instead of `rfid::Tag _tag`.

YannLocatelli (approved):
That's look pretty good!   
Except applying suggestion of @ladislas by setting Language in MagicCard struct, everything is good to me 👌

libs/RFIDKit/include/MagicCard.h
```diff
@@ -11,6 +11,12 @@
 
 namespace leka {
 
+enum class Language
+{
+	fr_FR = 1,
+	en_US = 2
+};
```
ladislas, line 18: Can you move this inside `struct MagicCard`?

ladislas (changes requested):
One last change :)

PR #1049, opened 2022-09-29
hugo/feature/Internationalize ActivityKit and RC

HPezz:
- [ ] Validated on robot

### DIFF
--- a/libs/ActivityKit/source/ActivityKit.cpp
+++ b/libs/ActivityKit/source/ActivityKit.cpp
@@ -34,6 +34,15 @@ void ActivityKit::stop()
 	_current_activity = nullptr;
 }
 
+void ActivityKit::displayMainMenu(const MagicCard &card)
+{
+	if (card.getLanguage() == MagicCard::Language::fr_FR) {
+		_videokit.displayImage("/fs/home/img/system/robot-misc-choose_activity-fr_FR.jpg");
+	} else {
+		_videokit.displayImage("/fs/home/img/system/robot-misc-choose_activity-en_US.jpg");
+	}
+}
+
 auto ActivityKit::isPlaying() const -> bool
 {
 	return _current_activity != nullptr;
--- a/libs/RobotKit/include/RobotController.h
+++ b/libs/RobotKit/include/RobotController.h
@@ -196,7 +196,9 @@ class RobotController : public interface::RobotController
 	void startAutonomousActivityMode() final
 	{
 		_lcd.turnOn();
-		_behaviorkit.displayAutonomousActivitiesPrompt();
+
+		auto card = _rfidkit.getLastMagicCardActivated();
+		_activitykit.displayMainMenu(card);
 	}
 
 	void stopAutonomousActivityMode() final
--- a/app/os/main.cpp
+++ b/app/os/main.cpp
@@ -373,7 +373,7 @@ namespace activities {
 
 }	// namespace activities
 
-auto activitykit = ActivityKit {};
+auto activitykit = ActivityKit {display::videokit};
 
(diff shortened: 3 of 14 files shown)


libs/RobotKit/include/RobotController.h
```diff
@@ -193,17 +193,7 @@ class RobotController : public interface::RobotController
...
-		_lcd.turnOn();
-		_behaviorkit.displayAutonomousActivitiesPrompt();
-	}
-
-	void stopAutonomousActivityMode() final
-	{
-		_behaviorkit.stop();
-		_activitykit.stop();
-	}
+	void stopAutonomousActivityMode() final { _activitykit.stop(); }
```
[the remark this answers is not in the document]
YannLocatelli: I mean, some instruction are for user-facing purpose, others are only for robot operations.

It is risky to rely on ActivityKit, a user-facing lib, for implementation of robot operations' purpose.

YannLocatelli (commented):
- [ ] Validated on the robot

libs/RobotKit/include/RobotController.h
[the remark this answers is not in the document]
YannLocatelli: **Suggestion.** Add `getLastMagicCardActivated` function in `RFIDKit` and use it on `startAutonomousActivityMode`

``` cpp
	void startAutonomousActivityMode() final
	{
		_lcd.turnOn();

        auto card = _rfidkit.getLastMagicCardActivated();
		_activitykit.displayMainMenu(card);
	}
```

libs/ActivityKit/source/ActivityKit.cpp
```diff
@@ -34,6 +34,15 @@ void ActivityKit::stop()
...
 }
 
+void ActivityKit::displayMainMenu(const MagicCard &card)
+{
+	if (card.getLanguage() == Language::fr_FR) {
+		_videokit.displayImage("/fs/home/img/system/robot-misc-choose_activity-fr_FR.jpg");
+	} else {
+		_videokit.displayImage("/fs/home/img/system/robot-misc-choose_activity-en_US.jpg");
+	}
+}
```
ladislas: shouldn't we check first if the card is indeed the dire_roll?

```suggestion
void ActivityKit::displayMainMenu(const MagicCard &card)
{
	if (card != MagicCard::dire_roll) {
		return;
	}
	
	if (card.getLanguage() == Language::fr_FR) {
		_videokit.displayImage("/fs/home/img/system/robot-misc-choose_activity-fr_FR.jpg");
	} else {
		_videokit.displayImage("/fs/home/img/system/robot-misc-choose_activity-en_US.jpg");
	}
}
```

libs/RobotKit/include/RobotController.h
```diff
@@ -196,7 +196,9 @@ class RobotController : public interface::RobotController
 	void startAutonomousActivityMode() final
 	{
 		_lcd.turnOn();
-		_behaviorkit.displayAutonomousActivitiesPrompt();
+
+		auto card = _rfidkit.getLastMagicCardActivated();
```
ladislas, line 200: this is smart but are you absolutely certain that it will work if:

- the user puts 2 cards (dice + something else)
- dice is read
- then the other one is read
- this function is called

?

ladislas (commented):
Still have two questions

libs/RobotKit/include/RobotController.h
HPezz: The execution speed between the dice_roll detection and the main menu display is so tiny that we couldn't put an other card immediately upon the dice roll detection, even willingly.

PR #1066, opened 2022-10-14
hugo/feature/Add IMUKit

HPezz:
Introduce IMUKit which compute angles from LSM6DSOX data through Mahony algorithm [[Link to Git repo inspiration](https://github.com/adafruit/Adafruit_AHRS/tree/master/src)]. 

Allows to get a correct image of Leka's orientation for approximatively 10 seconds until Earth rotation interferes too much.


libs/IMUKit/tests/IMUKit_test.cpp
```diff
@@ -0,0 +1,163 @@
...
+
+	EXPECT_CALL(mock_lsm6dsox, getData).WillRepeatedly(ReturnRef(data));
+
+	for (auto i = 0; i < 100; ++i) {
+		imukit.computeAngles();
+	}
+
+	auto [pitch, roll, yaw] = imukit.getAngles();
+
+	EXPECT_TRUE(pitch < -80 && pitch > -100);
```
YannLocatelli, line 75: Details but can be set this way to be more comfortable to read

```suggestion
	EXPECT_TRUE(-100 < pitch && pitch < -80);
```

libs/IMUKit/tests/IMUKit_test.cpp
```diff
@@ -0,0 +1,163 @@
...
+	for (auto i = 0; i < 100; ++i) {
+		imukit.computeAngles();
+	}
+
+	auto [pitch, roll, yaw] = imukit.getAngles();
+
+	EXPECT_EQ(pitch, 0);
+	EXPECT_EQ(roll, 0);
+	EXPECT_EQ(yaw, 180);
+}
```
YannLocatelli, line 163: Just to be sure, none of pitch/roll/yaw change, is that normal for an acceleration on Z axis?

libs/IMUKit/include/IMUKit.h
```diff
@@ -0,0 +1,53 @@
...
+namespace leka {
+
+class IMUKit
+{
+  public:
+	IMUKit(interface::EventLoop &event_loop, CoreAccelerometer &core_accelerometer, CoreGyroscope &core_gyroscope)
+		: _event_loop(event_loop),
+		  _core_accelerometer(core_accelerometer),
+		  _core_gyroscope(core_gyroscope) {
+
```
YannLocatelli, line 23:
```suggestion
    // nothing to do
```

tests/functional/tests/imu_kit/suite_imu_kit.cpp
```diff
@@ -0,0 +1,130 @@
...
+
+	"Initialization"_test = [&] {
+		expect(neq(&imukit, nullptr));
+		imukit.init();
+	};
+
+	"Start"_test = [&] {};
+
+	"get angles"_test = [&] {
+		imukit.start();
```
YannLocatelli, line 50: These tests are somehow not homogeneous, there are initialization, then some test then a scenario (that does not restart imukit). 

There is even a `start` before an `init` for imukit

Maybe you can move the `"get angles"` test in a scenario where you initialize imukit and start it, and `initialize`+`start` instead of `reset` in `scenario("imu - drift")`

libs/IMUKit/include/IMUKit.h
```diff
@@ -0,0 +1,53 @@
...
+		  };
+
+	struct Config {
+		const std::chrono::milliseconds delay {};
+		const float frequency {};
+	};
```
YannLocatelli, line 29: This struct `Config` is incomplete

- It is only used to instantiate a private member, therefore it does not have to be public
- If it means to be public, it should at least appears in parameter of one method (`ctor`, `init`, `start`)

YannLocatelli (commented):
One thing about struct `Config` used in IMUKit and some details~

libs/IMUKit/tests/IMUKit_test.cpp
HPezz: Indeed, with current configuration of IMU without magnetometer, yaw orientation is arbitrarily set at 180 in first measurement. While no angular speed is recorded by IMU, yaw will not change. I will add several tests with angular speed change to assure that yaw measurement isn't frozen. 

For pitch & roll, this is the remarkable point where roll and pitch are respectively equals to 0 or 180/0. Adding an epsilon on acceleration on X or Y axis makes roll fall to 179 for instance.

tests/functional/tests/imu_kit/suite_imu_kit.cpp
```diff
@@ -0,0 +1,130 @@
...
+
+				rtos::ThisThread::sleep_for(100ms);
+			}
+
+			expect(neq(i_batch, f_batch));
+		};
+	};
+
+	scenario("imu - drift") = [&] {
+		given("angles measurement is reset") = [&] {
```
ladislas, line 100:
```suggestion
		given("a new origin is set") = [&] {
```

see previous comment

tests/functional/tests/imu_kit/suite_imu_kit.cpp
```diff
@@ -0,0 +1,130 @@
...
+
+			then("I expect yaw to be reset to 180 degrees") = [&] {
+				auto [pitch, roll, yaw] = imukit.getAngles();
+				expect(yaw > default_min_bound_yaw)
+					<< "Yaw (" << yaw << ") lesser than minimal bound (" << default_min_bound_yaw << ")";
+				expect(yaw < default_max_bound_yaw)
+					<< "Yaw (" << yaw << ") greater than maximal bound (" << default_max_bound_yaw << ")";
+			};
+
+			then("I expect yaw to drift slightly") = [&] {
```
ladislas, line 111: when I wait for xxx seconds
then I expect yaw to drift

ladislas (changes requested):
That's a lot of great work! 👍 I tried to make to review as precise as possible and give the solutions when I have them.

YannLocatelli (approved):
Excellent! Thank you for you work 👏

libs/IMUKit/include/IMUKit.h
HPezz: I missed this review, my bad

spikes/lk_imu_kit/main.cpp
```diff
@@ -0,0 +1,56 @@
...
+	hello.start();
+
+	imu::internal::lsm6dsox.init();
+
+	imu::kit.init();
+	imu::kit.start();
+
+	while (true) {
+		auto [pitch, roll, yaw] = imu::kit.getAngles();
+		log_info("Pitch : %f, Roll : %f Yaw : %f", pitch, roll, yaw);
```
ladislas, line 52: cleaner output with padding for easier reading

is `%7.2f`:

- `7` is the total output length to add padding
- `.2` is the number of decimals

```suggestion
		log_info("Pitch : %7.2f, Roll : %7.2f Yaw : %7.2f", pitch, roll, yaw);
```

ladislas (approved):

PR #1069, opened 2022-10-17
ladislas+yann/feature/cmake generate os version

ladislas:
- :recycle: (version): Rename struct FirmwareVersion -> Version
- :wrench: (FirmwareKit): CMake - Generate os_version.h
- :wrench: (cmake-tools): Add OS_VERSION to config

replaces #1025

Needs: 

- [x] #1023
- [ ] Add workflow to check version are correspondant in config and generated binary
- [x] Rename `Firmware_Version` by `Version`
- [x] ~~Fix use of `settings.json`~~
- [x] Refactor getCurrentVersionFromFile

### DIFF
--- a/.vscode/settings.template.json
+++ b/.vscode/settings.template.json
@@ -80,6 +80,7 @@
 	"cmake.generator": "Ninja",
 	"cmake.configureSettings": {
 		"TARGET_BOARD": "LEKA_V1_2_DEV",
+		"OS_VERSION": "0.0.0",
 		"ENABLE_LOG_DEBUG": "ON",
 		"ENABLE_SYSTEM_STATS": "ON",
 		"MBED_PATH": "${workspaceFolder}/extern/mbed-os",
--- a/libs/FirmwareKit/CMakeLists.txt
+++ b/libs/FirmwareKit/CMakeLists.txt
@@ -4,9 +4,13 @@
 
 add_library(FirmwareKit STATIC)
 
+configure_file("${CMAKE_CURRENT_SOURCE_DIR}/include/os_version.h.in"
+	"${CMAKE_CURRENT_BINARY_DIR}/include/os_version.h")
+
 target_include_directories(FirmwareKit
 	PUBLIC
 	include
+	$<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}/include>
 )
 
 target_sources(FirmwareKit
--- a/libs/FirmwareKit/tests/FirmwareKit_test.cpp
+++ b/libs/FirmwareKit/tests/FirmwareKit_test.cpp
@@ -9,6 +9,7 @@
 #include "FirmwareKit.h"
 #include "gtest/gtest.h"
 #include "mocks/leka/FlashMemory.h"
+#include "os_version.h"
 
 using namespace leka;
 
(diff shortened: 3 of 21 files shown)


.vscode/settings.template.json
```diff
@@ -79,6 +79,7 @@
 	"cmake.generator": "Ninja",
 	"cmake.configureSettings": {
 		"TARGET_BOARD": "LEKA_V1_2_DEV",
+		"OS_VERSION": "1.2.3",
```
YannLocatelli, line 82: To avoid any confusion
```suggestion
		"OS_VERSION": "0.0.0",
```

libs/FirmwareKit/CMakeLists.txt
```diff
@@ -4,9 +4,13 @@
 
 add_library(FirmwareKit STATIC)
 
+configure_file("${CMAKE_CURRENT_SOURCE_DIR}/include/os_version.h.in"
+	"${CMAKE_CURRENT_BINARY_DIR}/include/os_version.h")
+
```
YannLocatelli, line 9: The `.in` file can be moved in another folder, for example

```suggestion
configure_file("${ROOT_DIR}/cmake/templates/LekaOS/os_version.h.in"
	"${CMAKE_CURRENT_BINARY_DIR}/include/os_version.h")
```

libs/FirmwareKit/tests/FirmwareKit_test.cpp
```diff
@@ -38,11 +39,14 @@ class FirmwareKitTest : public ::testing::Test
 	std::string bin_update_path				  = "/tmp/update-v2.0.0.bin";
 	std::array<uint8_t, 6> bin_update_content = {0x61, 0x62, 0x63, 0x64, 0x65, 0x66};	// "abcdef"
 
-	FirmwareVersion default_version = FirmwareVersion {1, 0, 0};
-	FirmwareVersion current_version = FirmwareVersion {1, 2, 3};
-	FirmwareVersion update_version	= FirmwareVersion {2, 0, 0};
+	// ? Default made sense when we were reading from a file in case the file was not readable
+	// ? With the #define OS_VERSION, this is not needed anymore
+	// TODO (@YannLocatelli): Remove the file related tests?
```
YannLocatelli, line 44: About this one, as mentioned in https://github.com/leka/LekaOS/pull/1025#issue-1373247002, 

> Refactor getCurrentVersionFromFile

A big part of FirmwareKit should be dropped, the part using SD. In same time, most of tests will be dropped too.

libs/FirmwareKit/tests/FirmwareKit_test.cpp
```diff
@@ -38,11 +39,14 @@ class FirmwareKitTest : public ::testing::Test
...
+	Version default_version = Version {1, 2, 0};
+	Version current_version = Version {1, 2, 0};
+	Version update_version	= Version {2, 0, 0};
```
YannLocatelli, line 47: `default_version` is used only when SD is not accessible. It is the lowest possible. As we stop to relate to SD, it is not anymore necessary.

Aside of this, `current_version` is an interesting one: it will need to be bumped too at each bump in that configuration. Do we need that?
Does the one that will be used in CI not sufficient?

YannLocatelli (commented):
- [ ] Validated on the robot

I can add the commit to clean up the whole `getCurrentVersionFromFile` part

libs/FirmwareKit/tests/FirmwareKit_test.cpp
ladislas: ah yes you're right! I've added back the `Needs` from #1025 to this PR description.

libs/FirmwareKit/tests/FirmwareKit_test.cpp
ladislas: > default_version is used only when SD is not accessible. It is the lowest possible. As we stop to relate to SD, it is not anymore necessary.

Yes we can remove it.

libs/FirmwareKit/CMakeLists.txt
YannLocatelli: Conclusion: `os_version` is only used by FirmwareKit, so it can stay here

libs/FirmwareKit/tests/FirmwareKit_test.cpp
YannLocatelli: See 

- #1072

YannLocatelli (approved):

PR #1086, opened 2022-10-26
:loud_sound: (functional): Busy sleep

MMyster:


tests/functional/include/tests/ft_utils.h
```diff
@@ -0,0 +1,21 @@
...
+
+#include "boost/ut.hpp"
+
+using namespace std::chrono;
+using namespace boost::ut;
+
+void busy_sleep(uint64_t milliseconds)
+{
+	for (uint64_t now = 0; now < milliseconds; ++now) {
+		rtos::ThisThread::sleep_for(1ms);
```
ladislas, line 17: do you need `1ms`, can't you make it 10 or 20? 

when you call `busy_sleep(100)` it means 100 dots will be printed, which is a lot. Even worst when you call is with `busy_sleep(2000)`

I think if `100ms` is okay, it would be better.

ladislas (changes requested):
First review. Thanks for the changes! :+1:

I have a few comments + suggestions.

tests/functional/include/tests/ft_utils.h
```diff
@@ -0,0 +1,21 @@
+// Leka - LekaOS
```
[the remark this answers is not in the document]
MMyster: But, I got some compilation problems...

tests/functional/include/tests/ft_utils.h
```diff
@@ -0,0 +1,21 @@
...
+using namespace boost::ut;
+
+void busy_sleep(uint64_t milliseconds)
+{
+	for (uint64_t now = 0; now < milliseconds; ++now) {
+		rtos::ThisThread::sleep_for(1ms);
+		log << ".";
+	}
+	log << "\n";
+}
```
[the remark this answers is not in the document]
MMyster: I would say that the last option seems the closest to reality

tests/functional/include/tests/ft_utils.h
MMyster: It does not bother but so the parameter will be a multiplier

ladislas: hmm, no it will not.

```cpp
void busy_sleep(std::chrono::milliseconds duration)
{
	static constexpr sleep_delay = 500ms;
	for (uint64_t now = 0; now < duration/sleep_delay; ++now) {
		rtos::ThisThread::sleep_for(sleep_delay);
	//...
```

ladislas:
closing PR for now as it's not needed anymore by #1087

PR #1087, opened 2022-10-26
Mmyster/feature/test functional file manager

MMyster:
Need : 

- [x] ~~#1086~~ not needed anymore

### DIFF
--- a/tests/functional/tests/file_manager/suite_file_manager.cpp
+++ b/tests/functional/tests/file_manager/suite_file_manager.cpp
@@ -1,0 +1,111 @@
+// Leka - LekaOS
+// Copyright 2022 APF France handicap
+// SPDX-License-Identifier: Apache-2.0
+
+#include <cstddef>
+
+#include "FileManagerKit.h"
+#include "boost/ut.hpp"
+#include "filesystem"
+#include "tests/config.h"
+#include "tests/utils.h"
+
+using namespace leka;
+using namespace boost::ut;
+
+struct path {
+	static inline const auto root = std::filesystem::path {"/fs/var/tmp/test"};
+
+	static inline const auto dir		  = root / "dir" / "";
+	static inline const auto sub_dir	  = root / "dir" / "sub";
+	static inline const auto dir_file	  = root / "dir" / "dir_file.txt";
+	static inline const auto sub_dir_file = root / "dir" / "sub" / "sub_dir_file.txt";
+
+	static inline const auto all = std::to_array({dir_file, sub_dir, dir, root});
+
+	static inline auto remove_all = []() {
+		log << "Cleaning up files, directories";
+
+		for (const auto &p: path::all) {
+			if (std::filesystem::exists(p)) {
+				log << "Removing:" << p;
+				std::filesystem::remove(p);
+				expect(not std::filesystem::exists(p)) << p << "still exists";
+			} else {
+				log << "Doesn't exit:" << p;
+			}
+		}
+		log << "\n";
+	};
+};
+
+suite suite_file_manager_kit = [] {
+	auto file		= FileManagerKit::File {};
+	auto input_data = std::to_array("Hello, Leka!");
+
+	utils::sd::init();
+	path::remove_all();
+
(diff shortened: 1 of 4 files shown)


tests/functional/include/tests/ft_utils.h
```diff
@@ -0,0 +1,21 @@
...
+#pragma once
+
+#include "rtos/ThisThread.h"
+
+#include "boost/ut.hpp"
+
+using namespace std::chrono;
+using namespace boost::ut;
+
+void busy_sleep(uint64_t milliseconds)
```
ladislas, line 14: So in the end I think this function is not useful anymore.

tests/functional/tests/file_manager/suite_file_manager.cpp
```diff
@@ -0,0 +1,78 @@
...
+
+void initializeSD()
+{
+	constexpr auto default_sd_blockdevice_frequency = uint64_t {25'000'000};
+
+	sd_blockdevice.init();
+	sd_blockdevice.frequency(default_sd_blockdevice_frequency);
+
+	fatfs.mount(&sd_blockdevice);
+}
```
ladislas, line 29: I think this should be moved to `utils.h` with the same structure as

https://github.com/leka/LekaOS/blob/ad37a1c39559572db0bf0f2603959bb99d2a9d2f/app/os/main.cpp#L92-L110

tests/functional/tests/file_manager/suite_file_manager.cpp
```diff
@@ -0,0 +1,78 @@
...
+
+suite suite_file_manager_kit = [] {
+	auto file = FileManagerKit::File {};
+
+	const std::filesystem::path tmp_dir			  = "/fs/var/tmp";
```
ladislas, line 34: problem was that you were trying to create `/fs/tmp` as seen here

[`93146f1` (#1087)](https://github.com/leka/LekaOS/pull/1087/commits/93146f1225cb09c9e315ae26a633c84267b83c68#diff-58867dbb0e58ede1dd8017cc825a323ab8fbe7e19686f1a44420be6748475082R34)

It seems like it's not possible to create a new dir at the root of the filesystem and one can only use already created to directories to create more.

good news `/fs/var/tmp` is available!

tests/functional/tests/file_manager/suite_file_manager.cpp
```diff
@@ -0,0 +1,78 @@
...
+	const std::filesystem::path test_sub_dir	  = test_dir / "sub/";
+	const std::filesystem::path test_filename	  = test_dir / "test_spike_lk_filesystem_kit";
+	const std::filesystem::path test_sub_filename = test_sub_dir / "test_spike_lk_filesystem_kit";
+	auto input_data								  = std::string {"Hello Leka"};
+
+	initializeSD();
+
+	std::filesystem::remove(test_filename);
+	std::filesystem::remove(test_sub_dir);
+	std::filesystem::remove(test_dir);
```
ladislas, line 45: order is important here.

I tried using `remove_all` but it was not working.

tests/functional/tests/file_manager/suite_file_manager.cpp
```diff
@@ -0,0 +1,78 @@
...
+		expect(open >> fatal) << "Failed to open file";
+	};
+
+	"write to new file "_test = [&] {
+		auto bytes = file.write(input_data);
+
+		expect(bytes == std::size(input_data) >> fatal) << "Failed to write file";
+	};
+
+	file.close();
```
ladislas, line 77: I would remove the directories+files again here, so you know you leave the fs in a sane state.

ladislas (changes requested):
It now works! 

image

tests/functional/include/tests/ft_utils.h
MMyster: Great if it works! The function may not be useful today but we can always think about it in the future

ladislas (approved):
Working now, ready to merge

YannLocatelli (approved):

PR #1088, opened 2022-10-26
:alembic: (stl): Test std::ranges support for arm-gcc + clang

ladislas:


spikes/stl_cxxsupport/main.cpp
```diff
@@ -94,5 +95,20 @@ auto main() -> int
 	auto back = span0.back();
 	log_info("span0.back() = %i", back);
 
+	log_info("Test std::range features");
+
+	constexpr auto range = [](const auto a, const auto b) {
+		auto x = std::ranges::views::iota(a) | std::ranges::views::take(b - a + 1);
+		return std::ranges::views::all(x);
+	};
```
YannLocatelli, line 103: I suggest to move it to `utils` and ensure to use the same for spike and functional test

spikes/stl_cxxsupport/main.cpp
```diff
@@ -94,5 +95,20 @@ auto main() -> int
...
+
+	for (auto x: range(1, 10)) {
+		log_info("x: %i", x);
+	}
```
YannLocatelli, line 107: This will prompt 

```
000:00:02.001 [INFO] [main.cpp:106] main > x: 1
000:00:02.001 [INFO] [main.cpp:106] main > x: 2
000:00:02.001 [INFO] [main.cpp:106] main > x: 3
000:00:02.001 [INFO] [main.cpp:106] main > x: 4
000:00:02.001 [INFO] [main.cpp:106] main > x: 5
000:00:02.001 [INFO] [main.cpp:106] main > x: 6
000:00:02.001 [INFO] [main.cpp:106] main > x: 7
000:00:02.001 [INFO] [main.cpp:106] main > x: 8
000:00:02.001 [INFO] [main.cpp:106] main > x: 9
000:00:02.001 [INFO] [main.cpp:106] main > x: 10
```

**Note.** For some usage of range, last value is excluded. Using math notation `range(a,b`) is equivalent to [a, b[.

It make sense for a for-loop of n values, range(n) should repeat n-times, from 0 to n-1.

spikes/stl_cxxsupport/main.cpp
```diff
@@ -94,5 +95,20 @@ auto main() -> int
 	auto back = span0.back();
 	log_info("span0.back() = %i", back);
 
+	log_info("Test std::range features");
+
+	constexpr auto range = [](const auto a, const auto b) {
+		auto x = std::ranges::views::iota(a) | std::ranges::views::take(b - a + 1);
+		return std::ranges::views::all(x);
```
YannLocatelli, line 102: Why don't use directly 

```suggestion
		return std::ranges::views::iota(a, b);
```

YannLocatelli: **Note Important.** Range go from a included to b excluded in that suggestion (  `[a, b[`  )

spikes/stl_cxxsupport/main.cpp
YannLocatelli: Can you explain a use case to figure out what is the best practice here? (Consider or not to include last value)

spikes/stl_cxxsupport/main.cpp
YannLocatelli: https://gcc.godbolt.org/z/7oc8bhqde

spikes/stl_cxxsupport/main.cpp
ladislas: can't do that yet as Apple's clang does not support std::ranges for now.
I'll look into using brew llvm's clang instead.

YannLocatelli:
Need 

- #1106

spikes/stl_cxxsupport/main.cpp
```diff
@@ -94,5 +96,30 @@ auto main() -> int
...
+	log_info("Test utils::ranges::sequences features");
+
+	for (auto val: utils::ranges::sequence({.start = 0, .end = 5})) {
+		log_info("x: %i", val);
+	}
+
+	for (auto val: utils::ranges::sequence({.start = 10, .end = 20})) {
+		log_info("y: %i", val);
+	}
+
```
YannLocatelli, line 123: Can add the other `sequence`

```suggestion
	for (auto val: utils::ranges::sequence({.start = 10, .end = 20})) {
		log_info("y: %i", val);
	}

	for (auto val: utils::ranges::sequence({.from = 101, .size = 10})) {
		log_info("z: %i", val);
	}

```

libs/Utils/tests/RangesUtils_test.cpp
```diff
@@ -0,0 +1,138 @@
...
+{
+	auto expected = std::to_array({10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20});
+
+	auto i = 0;
+	for (auto value: sequence({.from = 10, .size = 10})) {
+		EXPECT_EQ(value, expected.at(i));
+		++i;
+	}
+
+	EXPECT_EQ(i, 10) << "Loop ran " << i << " times, should run " << 10 << " times";
```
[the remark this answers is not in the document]
YannLocatelli:
```suggestion
	auto expected = std::to_array({10, 11, 12, 13, 14, 15, 16, 17, 18, 19});

	auto i = 0;
	for (auto value: sequence({.from = 10, .size = 10})) {
		EXPECT_EQ(value, expected.at(i));
		++i;
	}

	EXPECT_EQ(i, 10) << "Loop ran " << i << " times, should run " << 10 << " times";
```

libs/Utils/tests/RangesUtils_test.cpp
```diff
@@ -0,0 +1,138 @@
...
+{
+	auto expected = std::to_array({0});
+
+	auto i = 0;
+	for (auto value: sequence({.start = 0, .end = 0})) {
+		EXPECT_EQ(value, expected.at(i));
+		++i;
+	}
+
+	EXPECT_EQ(i, 1) << "Loop ran " << i << " times, should run " << 0 << " times";
```
YannLocatelli, line 46:
```suggestion
	EXPECT_EQ(i, 1) << "Loop ran " << i << " times, should run " << 0 + 1 << " times";
```

libs/Utils/tests/RangesUtils_test.cpp
```diff
@@ -0,0 +1,138 @@
...
+{
+	auto expected = std::to_array({1});
+
+	auto i = 0;
+	for (auto value: sequence({.start = 1, .end = 1})) {
+		EXPECT_EQ(value, expected.at(i));
+		++i;
+	}
+
+	EXPECT_EQ(i, 1) << "Loop ran " << i << " times, should run " << 0 << " times";
```
YannLocatelli, line 59:
```suggestion
	EXPECT_EQ(i, 1) << "Loop ran " << i << " times, should run " << 0 + 1 << " times";
```

YannLocatelli (commented):
Finally functional after `xcode-select --install`

Some details and how about CI?

YannLocatelli (approved):
Thank you~

PR #1100, opened 2022-10-31
:technologist: (make): Add target ut_lite for faster unit tests

ladislas:
Because of RobotKit, unit tests take an incredible amount of time to
compile and more specifically link the RobotKit static library

We add a config_unit_tests_lite target to build tests with only the
StateMachine tests and without the RobotController tests.


HPezz (approved):
- [ ] Validated on my computer

Lgtm

YannLocatelli (approved):

PR #1116, opened 2022-11-14
:zap: (LogKit): Process fifo by chuncks of 64 bytes

ladislas:
mbed question: https://forums.mbed.com/t/performance-question-bufferedserial-write-byte-after-byte-or-in-chuncks/18671


HPezz (approved):
LGTM

YannLocatelli (approved):

PR #1119, opened 2022-11-14
ladislas/feature/cmake python version

ladislas:
- :adhesive_bandage: (cmake): Add missing python shebang
- :hammer: (cmake): Pin python version to 3.10, brew's current default

### DIFF
--- a/cmake/scripts/check_python_package.py
+++ b/cmake/scripts/check_python_package.py
@@ -1,4 +1,7 @@
-# file which is invoked by the cmake build system to check if all necessary python packages are installed.
+#!/usr/bin/env python3
+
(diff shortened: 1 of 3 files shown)


HPezz (approved):
👍

cmake/scripts/check_python_package.py
```diff
@@ -1,4 +1,7 @@
-# file which is invoked by the cmake build system to check if all necessary python packages are installed.
+#!/usr/bin/env python3
```
YannLocatelli, line 1: This is also missing for `configure_cmake_for_target.py`, in the same `cmake/scripts` directory. Would you like to update it too?

PR #1142, opened 2022-11-18
:zap: (LogKit): Resize fifo to 8192 to allow prevent logs from disappearing

ladislas:


YannLocatelli (approved):
- [x] Validated on the robot

PR #1151, opened 2022-11-22
ladislas/bugfix/ut corejpeg missing include

ladislas:
Fix issue spotted at:

- https://github.com/leka/LekaOS/actions/runs/3527671293/jobs/5916981299#step:6:1435
- https://github.com/leka/LekaOS/actions/runs/3522366135/jobs/5905265031#step:6:1439

```
drivers/CoreVideo/CMakeFiles/CoreVideo.dir/source/CoreJPEG.cpp.o -c /home/runner/work/LekaOS/LekaOS/drivers/CoreVideo/source/CoreJPEG.cpp
/home/runner/work/LekaOS/LekaOS/drivers/CoreVideo/source/CoreJPEG.cpp:93:30: error: no member named 'adjacent_find' in namespace 'std'
                auto *find_index           = std::adjacent_find(std::begin(buffer), buffer_end_index, is_jpeg_soi_marker_predicate);
                                             ~~~~~^
1 error generated.
```

### DIFF
--- a/drivers/CoreVideo/include/CoreJPEG.hpp
+++ b/drivers/CoreVideo/include/CoreJPEG.hpp
@@ -4,9 +4,6 @@
 
 #pragma once
 
-#include <array>
-#include <cstdint>
-
 #include "JPEGImageProperties.hpp"
 #include "interface/JPEG.hpp"
 #include "interface/JPEGMode.hpp"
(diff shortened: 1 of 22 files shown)


YannLocatelli (changes requested):
It is better in my opinion to keep it the most explicit possible the origin of some used classe/interface/variable (e.g. `interface::STM32Hal`, `interface::File`, `JPEGImageProperties`, `jpeg::JPEG_SOI_MARKER`) and not popping from nowhere.

Same from standard libs, `#include <array>` is indeed not relevant in header file but it is used in source and might lead to unknown`std::array` warning/error

YannLocatelli (approved):

PR #1156, opened 2022-11-24
🚸 (update): Check file of requested update exists prior to update

YannLocatelli:
This will also avoid to reset system even file does not exists

### DIFF
--- a/libs/FirmwareKit/source/FirmwareKit.cpp
+++ b/libs/FirmwareKit/source/FirmwareKit.cpp
@@ -15,15 +15,25 @@ auto FirmwareKit::getCurrentVersion() -> Version
 	return Version {.major = semversion.major, .minor = semversion.minor, .revision = semversion.patch};
 }
 
-auto FirmwareKit::loadUpdate(const Version &version) -> bool
+auto FirmwareKit::getPathOfVersion(const Version &version) const -> std::filesystem::path
 {
-	auto path = std::array<char, 64> {};
-	snprintf(path.data(), std::size(path), _config.bin_path_format, version.major, version.minor, version.revision);
+	auto path_buffer = std::array<char, 64> {};
+	snprintf(path_buffer.data(), std::size(path_buffer), _config.bin_path_format, version.major, version.minor,
+			 version.revision);
 
-	return loadUpdate(path.data());
+	auto path = std::filesystem::path {path_buffer.begin()};
+
+	return path;
 }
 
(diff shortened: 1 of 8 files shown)


ladislas (approved):
LGTM 👍

HPezz (approved):
Nice work !

libs/FirmwareKit/source/FirmwareKit.cpp
```diff
@@ -15,15 +15,39 @@ auto FirmwareKit::getCurrentVersion() -> Version
...
+
+	auto file_exists						= false;
+	const auto k_minimal_expected_file_size = 300'000;
+	if (auto is_open = _file.open(path); is_open) {
+		file_exists = _file.size() >= k_minimal_expected_file_size;
+	}
+	_file.close();
+
+	return file_exists;
+}
```
ladislas, line 41:
```suggestion
auto FirmwareKit::isVersionAvailable(const Version &version) -> bool
{
	auto path = getPathOfVersion(version);
	auto file_exists = false;
	
	if (auto is_open = _file.open(path); is_open) {
		constexpr auto kMinimalFileSizeInBytes = std::size_t {300'000};
		file_exists = _file.size() >= kMinimalFileSizeInBytes;
	}
	
	_file.close();

	return file_exists;
}
```

libs/FirmwareKit/tests/FirmwareKit_test.cpp
```diff
@@ -73,23 +52,60 @@ TEST_F(FirmwareKitTest, getCurrentVersion)
...
+
+	auto is_version_available = firmwarekit.isVersionAvailable(unexisting_version);
+
+	EXPECT_FALSE(is_version_available);
+}
+
+TEST_F(FirmwareKitTest, isVersionAvailableFileTooSmall)
+{
+	auto dummy_version				   = Version {99, 99, 9999};
+	std::string bin_dummy_version_path = "fs/usr/os/LekaOS-99.99.9999.bin";
```
ladislas, line 97: the test is great, but I'd prefer we don't use `fs/` for creating files.

if for some reason the test crashes before removing at the end, you'll end up with new files in `fs/` and there is a risk someone might commit those by accident.

It would be better to use the `tmp/` directory for this case, even though it means instantiating a new firmwarekit object

ladislas (changes requested):
Final few tweaks and we're good to go 👍

PR #1163, opened 2022-11-25
Yann/feature/filemanagerkit/add exists

YannLocatelli:
# Validations

- [x] bootloader - get Counter
- [x] spike_lk_config_kit (existing and non existing)
- [x] firmware - update (isVersionAvailable + load)
- [x] spike_lk_lcd
- [x] spike_lk_audio

# Requirements

- [x] #1281

### DIFF
--- a/app/bootloader/main.cpp
+++ b/app/bootloader/main.cpp
@@ -112,11 +112,11 @@ namespace factory_reset {
 
 	auto getCounter() -> uint8_t
 	{
-		FileManagerKit::File file {internal::factory_reset_counter_path, "r"};
-
-		if (!file.is_open()) {
+		if (FileManagerKit::file_is_missing(internal::factory_reset_counter_path)) {
 			return default_limit + 1;
 		}
+
+		auto file = FileManagerKit::File {internal::factory_reset_counter_path, "r"};
 
 		auto data = std::array<uint8_t, 1> {};
 		file.read(data);
--- a/libs/FileManagerKit/source/FileManagerKit.cpp
+++ b/libs/FileManagerKit/source/FileManagerKit.cpp
@@ -20,3 +20,14 @@ auto FileManagerKit::remove(const std::filesystem::path &path) -> bool
 {
 	return std::filesystem::remove(path);
 }
+
+auto FileManagerKit::file_exists(const std::filesystem::path &path) -> bool
+{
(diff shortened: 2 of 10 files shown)


YannLocatelli:
Regarding of the solutions, is it still relevant?

This PR is paused for now

libs/FileManagerKit/source/FileManagerKit.cpp
```diff
@@ -20,3 +20,9 @@ auto FileManagerKit::remove(const std::filesystem::path &path) -> bool
 {
 	return std::filesystem::remove(path);
 }
+
+auto FileManagerKit::file_exists(const std::filesystem::path &path) -> bool
+{
+	auto file = FileManagerKit::File {path, "r"};
+	return file.is_open();
```
ladislas, line 27: have you tested that the file is really closed when it goes out of scope?

app/bootloader/main.cpp
```diff
@@ -112,12 +112,13 @@ namespace factory_reset {
 
 	auto getCounter() -> uint8_t
 	{
-		FileManagerKit::File file {internal::factory_reset_counter_path, "r"};
-
-		if (!file.is_open()) {
+		if (auto file_does_not_exist = !FileManagerKit::file_exists(internal::factory_reset_counter_path);
+			file_does_not_exist) {
```
ladislas, line 116: for this one and for all the others, I'm wondering if the `if-statement` is really easier to read with the `init-statement`

`init-statement` are useful when you need the variable inside the `if` clause without the need of an external variable:

```cpp
// without
auto condition = some_function();

if (condition > 10) {
	do_something_with(condition);
}

// with
if (auto condition = some_function(); condition > 10) {
	do_something_with(condition);
}
```

if our case we don't do anything with the variable.

furthermore, a lot of times, we don't want to know `if the file DOES exist` but `if the file DOES *NOT* exist`

In this case, it make sens to provide the user with a function to do just that:

```cpp
auto file_exists(path) -> bool {
	return ...
}

auto file_does_not_exist(path) -> bool {
	return !file_exists(path);
}
```
that can be used:

```cpp
if (file_does_not_exist(path)) {
	// do something
}
```

ladislas (approved):
LGTM :+1:

You can fixup and I'll merge asap :)

PR #1171, opened 2022-11-30
:recycle: (tools): Set response timeout to 30s

HPezz:
Increasing the "no-response delay" to allow longer functional tests. 

PR created on the following context :  Measurement of the IMU's yaw drift for 10 seconds, meaning amount of time during the drift should be lower than a degree


ladislas (approved):

YannLocatelli (commented):
Can you add motivation in description + the impact on tests, for example duration of tests

I add a screenshot for information, does it really double time of test or is it ponctual
Screenshot 2022-11-30 at 2 59 41 PM

YannLocatelli (approved):

PR #1181, opened 2022-12-02
:bug: (CoreMotor): Resume pwm before setting up speed

HPezz:


drivers/CoreMotor/source/CoreMotor.cpp
```diff
@@ -33,6 +33,7 @@ void CoreMotor::setDirections(int dir_1, int dir_2)
 
 void CoreMotor::setSpeed(float speed)
 {
+	_speed.resume();
```
ladislas, line 36: line break after
```suggestion
	_speed.resume();
	
```

ladislas (changes requested):
LGTM 👍 just missing line break

YannLocatelli (approved):

PR #1185, opened 2022-12-05
:construction_worker: (config): ut + sanitizers - use gcc 11 as experimental

ladislas:
This allows workflow to continue on error


ladislas:
goal is to move to gcc@12 directly, closing the issue for now

PR #1192, opened 2022-12-08
:children_crossing: (emergency stop): Add minimal delay to use emergency stop

HPezz:


libs/RobotKit/include/RobotController.h
```diff
@@ -339,7 +339,7 @@ class RobotController : public interface::RobotController
 		auto is_autonomous_mode		= state_machine.is(system::robot::sm::state::autonomous_activities);
 		auto NOT_is_autonomous_mode = !is_autonomous_mode;
 
-		if (card == MagicCard::emergency_stop) {
+		if (card == MagicCard::emergency_stop && rtos::Kernel::Clock::now() - init_timestamp > 10s) {
```
ladislas, line 342: separate it in two checks as they are not relevant at the same time:

```suggestion
		if (rtos::Kernel::Clock::now() - init_timestamp < 10s) {
			return;
		}
		
		if (card == MagicCard::emergency_stop) {
```

libs/RobotKit/include/RobotController.h
```diff
@@ -468,6 +468,8 @@ class RobotController : public interface::RobotController
 	std::chrono::seconds _idle_timeout_duration {600};
 	interface::Timeout &_timeout;
 
+	rtos::Kernel::Clock::time_point init_timestamp = rtos::Kernel::Clock::now();
```
ladislas, line 471: can be made `const`

```suggestion
	const rtos::Kernel::Clock::time_point kSystemStartupTimestamp = rtos::Kernel::Clock::now();
```

ladislas (commented):
LGTM 👍

ladislas (approved):

ladislas (approved):
LGMT 👍

libs/RobotKit/tests/RobotController_test_stateIdle.cpp
```diff
@@ -161,7 +176,7 @@ TEST_F(RobotControllerTest, stateIdleDiceRollDetectedDelayOverEventAutonomousAct
 {
 	rc.state_machine.set_current_states(lksm::state::idle);
 
-	auto minimal_delay_over = 1001ms;
+	auto minimal_delay_over = 1001s;
```
YannLocatelli, line 179: Why this change from `ms` to `s` ?

If you want a delay over 11s, it is just `11s` or `11001ms`

libs/RobotKit/include/RobotController.h
```diff
@@ -345,7 +345,7 @@ class RobotController : public interface::RobotController
 		auto is_autonomous_mode		= state_machine.is(system::robot::sm::state::autonomous_activities);
 		auto NOT_is_autonomous_mode = !is_autonomous_mode;
 
-		if (card == MagicCard::emergency_stop) {
+		if (card == MagicCard::emergency_stop && rtos::Kernel::Clock::now() - kSystemStartupTimestamp > 10s) {
```
YannLocatelli, line 348: You don't want a parameter as the startup duration to be drown in the code, I suggest to set at the start of the method or set in member variables.

YannLocatelli (changes requested):
A question about a delay, and a suggestion to move the delay to be more accessible.

PR #1195, opened 2022-12-08
🐛 (rc): Move the long SHA256 get process to the RC's EventQueue

YannLocatelli:


HPezz (approved):
Nice catch

ladislas (approved):
LGTM 👍 

thinking about it, the issue with sha256 might be same as the issue with file exchange: when reading from sd card, it took too much time for BLE to process/send the data and it caused the crash.

PR #1197, opened 2022-12-09
✨ (file/filemanager): Add clear file

YannLocatelli:


libs/FileManagerKit/source/File.cpp
```diff
@@ -33,12 +33,14 @@ File::File(const std::filesystem::path &path, const char *mode)
...
 auto File::open(const char *path, const char *mode) -> bool
 {
+	_path = path;
 	_file.reset(std::fopen(path, mode));
 	return is_open();
 }
 
 auto File::open(const std::filesystem::path &path, const char *mode) -> bool
 {
+	_path = path;
```
ladislas, line 43: same for those two changes, part of different commit adding _path

libs/FileManagerKit/include/FileManagerKit.h
```diff
@@ -57,8 +60,10 @@ struct File : public interface::File, public mbed::NonCopyable<File> {
 
   private:
 	std::unique_ptr<FILE, decltype(&fclose)> _file {nullptr, &fclose};
+	std::filesystem::path _path;
```
ladislas, line 63: could you make this change part of a different commit?

libs/FileManagerKit/tests/File_test.cpp
```diff
@@ -864,3 +864,30 @@ TEST_F(FileTest, clearError)
...
+
+TEST_F(FileTest, clear)
+{
+	auto input_data = std::to_array<uint8_t>({0x61, 0x62, 0x63, 0x64, 0x65, 0x66});	  // "abcdef"
+
+	writeTempFile(input_data);
+
+	file.open(tempFilename);
+	auto size = file.size();
+	EXPECT_NE(0, size);
```
ladislas, line 876:
```suggestion
	EXPECT_NE(size, 0);
	EXPECT_EQ(size, std::size(input_data);
```

libs/FileManagerKit/tests/File_test.cpp
```diff
@@ -864,3 +864,30 @@ TEST_F(FileTest, clearError)
...
+
+	file.clear();
+
+	size = file.size();
+	EXPECT_EQ(0, size);
```
ladislas, line 881:
```suggestion
	EXPECT_EQ(size, 0);
```

libs/FileManagerKit/source/FileManagerKit.cpp
```diff
@@ -20,3 +20,10 @@ auto FileManagerKit::remove(const std::filesystem::path &path) -> bool
 {
 	return std::filesystem::remove(path);
 }
+
+void FileManagerKit::clear(const std::filesystem::path &path)
+{
+	if (std::filesystem::exists(path)) {
```
ladislas, line 26: this will suffer from the same issue we had in the other PR, bringing a lot of STL code.

I would prefer we don't add it in this PR, especially without adding a use case in the spike or functional test to monitor the impact.

ladislas: or use the same workaround you used in the other PR.

libs/FileManagerKit/tests/FileManagerKit_test.cpp
```diff
@@ -81,3 +81,30 @@ TEST_F(FileSystemTest, removeNotExistingDirectory)
...
+
+	auto initial_file_size	= std::filesystem::file_size(path_A);
+	auto initial_free_space = std::filesystem::space(path_A).free;
+
+	EXPECT_NE(initial_file_size, 0);
+
+	FileManagerKit::clear(path_A);
+
+	auto final_file_size  = std::filesystem::file_size(path_A);
+	auto final_free_space = std::filesystem::space(path_A).free;
```
ladislas, line 97: I'm not sure I understand why you want to test it like this.

why not just write to the file, measure the size, clear it, measure again?

ladislas (changes requested):
the File part LGTM w/ small suggestions 👍 

regarding the FileManager part, I would prefer we keep it out of this PR as it's using `exists` which will suffer from the same issues we previously had.

the FileManager test part is also a bit strange, I'm not sure I understand why you are testing like this.

libs/FileManagerKit/tests/FileManagerKit_test.cpp
YannLocatelli: It's to verify that it is not just a truncation of a file: move the EOF character to the beginning of file and kee the previous data allocated but inaccessible.

ladislas (approved):

PR #1199, opened 2022-12-12
hugo/fix/Refactor Emergency stop initial delay

HPezz:
- :white_check_mark: (RC tests): Fix seconds/millisecond issue in RC test
- :recycle: (RC): Use a variable for emergency stop init delay


libs/RobotKit/include/RobotController.h
```diff
@@ -340,12 +340,13 @@ class RobotController : public interface::RobotController
...
-		auto is_autonomous_mode		= state_machine.is(system::robot::sm::state::autonomous_activities);
-		auto NOT_is_autonomous_mode = !is_autonomous_mode;
+		auto is_playing					   = _activitykit.isPlaying();
+		auto NOT_is_playing				   = !is_playing;
+		auto is_autonomous_mode			   = state_machine.is(system::robot::sm::state::autonomous_activities);
+		auto NOT_is_autonomous_mode		   = !is_autonomous_mode;
+		const auto kTimeSinceSystemStartup = rtos::Kernel::Clock::now();
 
-		if (card == MagicCard::emergency_stop && rtos::Kernel::Clock::now() - kSystemStartupTimestamp > 10s) {
+		if (card == MagicCard::emergency_stop && kTimeSinceSystemStartup - kSystemStartupTimestamp > 10s) {
```
ladislas, line 349: I don't think that's what @YannLocatelli meant.

more something like this:

```suggestion
		const auto startup_delay_elapsed = rtos::Kernel::Clock::now() - kSystemStartupTimestamp > 10s;

		if (card == MagicCard::emergency_stop && startup_delay_elapsed) {
```

HPezz: :+1:

PR #1215, opened 2022-12-15
hugo/feature/Improve MotionKit

HPezz:
Improve smart mobile reinforcer 1080 rotation and add timeout to Motion Kit to prevent infinite rotation

- [x] Validated on robot

### DIFF
--- a/libs/MotionKit/include/MotionKit.h
+++ b/libs/MotionKit/include/MotionKit.h
@@ -16,8 +16,14 @@ class MotionKit
 {
   public:
 	MotionKit(interface::Motor &motor_left, interface::Motor &motor_right, IMUKit &imu_kit,
-			  interface::EventLoop &event_loop)
-		: _motor_left(motor_left), _motor_right(motor_right), _imukit(imu_kit), _event_loop(event_loop) {};
+			  interface::EventLoop &event_loop, interface::Timeout &timeout)
+		: _motor_left(motor_left),
+		  _motor_right(motor_right),
+		  _imukit(imu_kit),
+		  _event_loop(event_loop),
+		  _timeout(timeout)
+	{
+	}
 
 	void init();
 
@@ -37,6 +43,7 @@ class MotionKit
 	interface::Motor &_motor_right;
 	IMUKit &_imukit;
 	interface::EventLoop &_event_loop;
+	interface::Timeout &_timeout;
 	PID _pid;
 
 	uint8_t _rotations_to_execute = 0;
--- a/libs/MotionKit/include/MotionKit.h
+++ b/libs/MotionKit/include/MotionKit.h
@@ -46,10 +46,16 @@ class MotionKit
 	bool _stabilisation_requested  = false;
 	bool _rotate_x_turns_requested = false;
 
-	const float kReferenceAngle		   = 180.F;
-	const float kMinimalViableRobotPwm = 0.25F;	  // ? Under this pwm value, torque is too low to spin the wheel
+	const float kReferenceAngle = 180.F;
+	const float kPIDMaxValue	= 1.8F;
+
+	// ? When the motor is stopped, PWM values under kMinimalViableRobotPwm are too low to generate enough torque for
+	// ? the motor to start spinning ? At the same time, kMinimalViableRobotPwm needs to be the lowest possible to avoid
+	// ? overshooting when the target is reached
+
+	const float kMinimalViableRobotPwm = 0.15F;
 	const float kPwmMaxValue		   = 1.F;
-	const float kPwmMarginLimit		   = 0.28F;
+	const float kEpsilon			   = 0.02F;
 };
 
 }	// namespace leka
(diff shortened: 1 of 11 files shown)


libs/MotionKit/include/MotionKit.h
```diff
@@ -16,8 +16,12 @@ class MotionKit
...
   public:
 	MotionKit(interface::Motor &motor_left, interface::Motor &motor_right, IMUKit &imu_kit,
-			  interface::EventLoop &event_loop)
-		: _motor_left(motor_left), _motor_right(motor_right), _imukit(imu_kit), _event_loop(event_loop) {};
+			  interface::EventLoop &event_loop, interface::Timeout &timeout)
+		: _motor_left(motor_left),
+		  _motor_right(motor_right),
+		  _imukit(imu_kit),
+		  _event_loop(event_loop),
+		  _timeout(timeout) {};
```
ladislas, line 24: having a `;` is an error

```suggestion
		  _timeout(timeout) {}
```

libs/MotionKit/include/MotionKit.h
```diff
@@ -47,9 +52,9 @@ class MotionKit
 	bool _rotate_x_turns_requested = false;
 
 	const float kReferenceAngle		   = 180.F;
-	const float kMinimalViableRobotPwm = 0.25F;	  // ? Under this pwm value, torque is too low to spin the wheel
+	const float kMinimalViableRobotPwm = 0.15F;	  // ? Under this pwm value, torque is too low to spin the wheel
 	const float kPwmMaxValue		   = 1.F;
-	const float kPwmMarginLimit		   = 0.28F;
+	const float kPwmMarginLimit		   = 0.17F;
```
ladislas, line 57: why those values? what changed?

libs/MotionKit/source/MotionKit.cpp
```diff
@@ -106,7 +111,7 @@ void MotionKit::run()
 
 auto MotionKit::mapSpeed(float speed) const -> float
 {
-	return utils::math::map(speed, 0.F, 1.F, kMinimalViableRobotPwm, kPwmMaxValue);
+	return utils::math::map(speed, 0.F, 1.8F, kMinimalViableRobotPwm, kPwmMaxValue);
```
ladislas, line 114: why this change? you'll get PWM values > 1.0, no?

ladislas (changes requested):
looks good, but I have a few questions on the values changed.

has this been tested on a closed robot -- I mean **really** closed

libs/MotionKit/source/MotionKit.cpp
YannLocatelli: Oh, you should set it to the maximum value of PID if it is instead of magic number `1.8F`

libs/MotionKit/include/MotionKit.h
YannLocatelli: Yes, why this is more performant than last time? Because the robot is closed?

libs/MotionKit/source/MotionKit.cpp
```diff
@@ -43,6 +43,11 @@ void MotionKit::rotate(uint8_t number_of_rotations, Rotation direction,
 	_motor_left.spin(direction, kPwmMaxValue);
 	_motor_right.spin(direction, kPwmMaxValue);
 
+	auto on_timeout = [this] { stop(); };
+
+	_timeout.onTimeout(on_timeout);
+	_timeout.start(10s);
```
YannLocatelli, line 49: `_timeout.stop()` in `MotionKit::stop` too. Otherwise it motion will be too short for two Reinforcer in a row.

libs/MotionKit/tests/MotionKit_test.cpp
```diff
@@ -36,12 +37,14 @@ class MotionKitTest : public ::testing::Test
 	mock::Accelerometer accel {};
 	mock::Gyroscope gyro {};
 
+	mock::Timeout mock_timeout {};
```
YannLocatelli, line 40: Add a test using the timeout callback to ensure you stop the motion at end of the timeout

YannLocatelli (changes requested):

libs/MotionKit/include/MotionKit.h
```diff
@@ -51,10 +53,15 @@ class MotionKit
 	bool _stabilisation_requested  = false;
 	bool _rotate_x_turns_requested = false;
 
-	const float kReferenceAngle		   = 180.F;
-	const float kMinimalViableRobotPwm = 0.15F;	  // ? Under this pwm value, torque is too low to spin the wheel
+	const float kReferenceAngle = 180.F;
+	const float kPIDMaxValue	= 1.8F;
+
+	// ? Under this pwm value, torque is too low to spin the wheel without inertia.
+	// ? This value needs to be the lowest possible to prevent inertia drift after reaching target
```
ladislas, line 60:
```suggestion
	// ? When the motor is stopped, PWM values under kMinimalViableRobotPwm are too low to generate enough torque for the motor to start spinning
	// ? At the same time, kMinimalViableRobotPwm needs to be the lowest possible to avoid overshooting when the target is reached
```

YannLocatelli (approved):

ladislas (approved):

PR #1216, opened 2022-12-15
🚸 (rc): Update charge status same time as battery level

YannLocatelli:
Fix charging status not updated in FileExchange state


ladislas (approved):
LGTM 👍

PR #1218, opened 2022-12-15
hugo/feature/Change animation blinkOnCharge

HPezz:
- :children_crossing: (MotionKit): Improve smart reinforcer performance
- :recycle: (MotionKit): Add timeout to rotate() function
- :recycle: (libs): Update Motionkit calls
- :children_crossing: (BlinkOnCharge): Improve animation
- :children_crossing: (BehaviorKit): Trigger back BlinkOnCharge after BleConnection

### DIFF
--- a/libs/LedKit/include/animations/BlinkOnCharge.h
+++ b/libs/LedKit/include/animations/BlinkOnCharge.h
@@ -2,21 +2,27 @@
 // Copyright 2022 APF France handicap
 // SPDX-License-Identifier: Apache-2.0
 
-//    0  +               1            + Loop
-//       |                            |
-//       |                            |
-//  -----|                            |- Ear light on
-//       |                            |
-//       |                            |
-//       |                            |
-//       |                            |
-//       |                            |
-//       |                            |
-//       |                            |
-//       |                            |
-//       |                            |
-//       |----------------------------| - Black
-//       |                            |
+// +        0        +         1        +                      2	            		+
+// |                 |\                 |  												|	-- White
+// |                /| \                |												|
+// |               / |  \               |                                               |
+// |              /  |   \              |                                               |
+// |             /   |    \             |                                               |
+// |            /    |     \            |												|
+// |           /     |      \           |												|
+// |          /      |       \          |												|
+// |         /       |        \         |												|
+// |        /        |         \        |												|
+// |       /         |          \       |												|
+// |      /          |           \      |												|
+// |     /           |            \     |												|
+// |    /            |             \    |												|
+// |   /             |              \   |												|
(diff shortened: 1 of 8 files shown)


libs/LedKit/include/animations/BlinkOnCharge.h
```diff
@@ -45,6 +45,13 @@ class BlinkOnCharge : public interface::LEDAnimation
 	uint8_t _step		  = 0;
 	uint8_t _stage		  = 0;
 
+	void stage0();
+	void stage1();
+	void stage2();
```
ladislas, line 50: can you name them something meaningful as you only have 3 and they do specific things

ladislas (changes requested):
LGTM besides stage naming 👍

libs/BehaviorKit/source/BehaviorKit.cpp
```diff
@@ -84,6 +84,9 @@ void BehaviorKit::bleConnection(bool with_video)
 	_ledkit.start(&led::animation::ble_connection);
 	if (with_video) {
 		_videokit.playVideoOnce("/fs/home/vid/system/robot-system-ble_connection-wink-no_eyebrows.avi");
+	} else {
+		rtos::ThisThread::sleep_for(5s);
+		blinkOnCharge();
```
YannLocatelli, line 89: This function (bleConnection) does not have notion of charge, only of video.

This workaround is fine but bring some confusion in BehaviorKit implementation.

YannLocatelli (changes requested):
Adding a charge notion in bleConnection is not good, and I just noticed that tests mention a condition on state connection or not: charging

State machine is composite, in the ideology to differentiate connection and robot states. So there are kind of not related.

Moreover Behaviors should not be aware of the state.

--

I don't have solution yet, I need some time to think about it.

libs/BehaviorKit/tests/BehaviorKit_test.cpp
```diff
@@ -109,7 +109,6 @@ TEST_F(BehaviorKitTest, bleConnectionWhileCharging)
 TEST_F(BehaviorKitTest, bleConnectionWhileNotCharging)
 {
 	EXPECT_CALL(mock_videokit, playVideoOnce);
-	EXPECT_CALL(mock_ledkit, start(isSameAnimation(&led::animation::ble_connection))).Times(1);
```
YannLocatelli, line 112: Keep this one

YannLocatelli (changes requested):
A typo + missing case

YannLocatelli (approved):

ladislas (approved):

PR #1220, opened 2022-12-16
🚸 (rc): Display an image during FileExchange

YannLocatelli:
Charging status is considered

### DIFF
--- a/libs/BehaviorKit/source/BehaviorKit.cpp
+++ b/libs/BehaviorKit/source/BehaviorKit.cpp
@@ -92,6 +92,11 @@ void BehaviorKit::working()
 	_videokit.displayImage("/fs/home/img/system/robot-face-smiling-slightly.jpg");
 }
 
+void BehaviorKit::fileExchange()
+{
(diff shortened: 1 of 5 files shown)


libs/BehaviorKit/source/BehaviorKit.cpp
```diff
@@ -92,6 +92,11 @@ void BehaviorKit::working()
 	_videokit.displayImage("/fs/home/img/system/robot-face-smiling-slightly.jpg");
 }
 
+void BehaviorKit::fileExchange()
+{
+	_videokit.displayImage("/fs/home/img/system/robot-file-exchange-reception.jpg");
```
YannLocatelli, line 97: @ladislas the name of file to plan is here

ladislas, line 97: let's start with that:

```suggestion
	_videokit.displayImage("/fs/home/img/system/robot-file_exchange.jpg");
```

I'll make something generic after the release

ladislas (approved):
LGTM, thanks @YannLocatelli ! 

I've added the name of the file to use

PR #1221, opened 2022-12-16
🚚 (behavior): Rename misleading bleConnection tests

YannLocatelli:
bleConnection don't have charging parameter


ladislas (approved):
renaming LGTM 👍 

but the `behaviorkit.bleConnection(false);` was not clear to me, I had to go read the code to understand what was `false`

as with `playVideo` it would be better to have both `bleConnectionWithVideo` and `bleConnectionWithoutVideo`

PR #1231, opened 2022-12-22
:hammer: (cmake): FirmwareKit - depend on os_version file

ladislas:
This commit make FirmwareKit (and directory) depend on os_version config
file

This means that when the version is updated, cmake configure is run
again and the os_version.h file is regenerated with the new version

This should make development as build number could be added in the
future as well


libs/FirmwareKit/tests/FirmwareKit_test.cpp
```diff
@@ -26,9 +27,11 @@ class FirmwareKitTest : public ::testing::Test
...
 	// void TearDown() override {}
 
-	Version current_version = Version {1, 2, 0};
-
-	std::string current_version_str = OS_VERSION;
+	Version current_version = Version {
+		semver::version {OS_VERSION}.major,
+		semver::version {OS_VERSION}.minor,
+		semver::version {OS_VERSION}.patch,
+	};
```
YannLocatelli, line 34: This is in conflict with issue

- https://github.com/leka/LekaOS/issues/1071

tests/functional/tests/firmware_kit/suite_firmware_kit.cpp
```diff
@@ -0,0 +1,36 @@
...
+using namespace std::chrono;
+
+auto qspi		   = CoreQSPI();
+auto flash_manager = CoreFlashManagerIS25LP016D(qspi);
+auto flash_memory  = CoreFlashIS25LP016D(qspi, flash_manager);
+
+suite suite_firmware_kit = [] {
+	"config version and firmware kit version are the same"_test = [&] {
+		auto firmwarekit		  = FirmwareKit(flash_memory, FirmwareKit::DEFAULT_CONFIG);
+		auto firmware_kit_version = firmwarekit.getCurrentVersion();
```
YannLocatelli, line 28: Homogeneous 
```suggestion
		auto firmwarekit		  = FirmwareKit(flash_memory, FirmwareKit::DEFAULT_CONFIG);
		auto firmwarekit_version = firmwarekit.getCurrentVersion();
```

YannLocatelli (changes requested):
Decide if the project depend or not on semver

libs/FirmwareKit/tests/FirmwareKit_test.cpp
ladislas: yes and no -- the idea of removing semver was to remove complicated code just to get the version number.

semver is using `constexpr` almost everywhere, which means that the actual work is done at compile time and not run time, so the final binary should only have the actual major, minor and patch values.

this must be further investigated to be confirmed.

on the other hand, removing semver will mean finding another way to parse the OS_VERSION, either by using lines for each values or different files. A bit more complicated to handle cleanly in cmake.

also, semver is already being used in embedded code, so using it in unit tests is normal and cleans up things while we investigate further and decide to keep it or remove it for something else.

YannLocatelli (approved):

PR #1234, opened 2023-01-01
:alembic: (CoreLed): Check the impact of using consteval instead of constexpr

ladislas:


ladislas:
closing as not necessary

PR #1238, opened 2023-01-02
♻️ (rc): Separate timeout: inner state and state transition

YannLocatelli:

PR #1239, opened 2023-01-02
✨ (rc): Add DeepSleep state

YannLocatelli:
## Requirements

- [x] #1238

### DIFF
--- a/libs/RobotKit/include/RobotController.h
+++ b/libs/RobotKit/include/RobotController.h
@@ -305,6 +305,8 @@ class RobotController : public interface::RobotController
 		stopActuators();
 	}
 
+	void suspendHardwareForDeepSleep() final { log_info("TO IMPLEMENT - configuring hardware for deep sleep"); }
+
 	void resetEmergencyStopCounter() final { _emergency_stop_counter = 0; }
 
 	void raise(auto event)
--- a/libs/RobotKit/include/RobotController.h
+++ b/libs/RobotKit/include/RobotController.h
@@ -90,6 +90,17 @@ class RobotController : public interface::RobotController
 
 	void stopSleepTimeout() final { _timeout_state_transition.stop(); }
 
+	void startDeepSleepTimeout() final
+	{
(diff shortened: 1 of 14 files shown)


HPezz (approved):
Nice work 👍

libs/RobotKit/include/RobotController.h
```diff
@@ -294,6 +305,11 @@ class RobotController : public interface::RobotController
 		stopActuators();
 	}
 
+	void turnOffHardware() final
```
ladislas, line 308:
```suggestion
	void suspendHardwareForDeepSleep() final
	{
		log_info("Entering deep sleep");
	}
```

ladislas (approved):
LGTM 👍 
small naming suggestion as well

libs/RobotKit/include/RobotController.h
YannLocatelli: Note. logs is not necessary since SM already display it

ladislas: yes I know, but SM displays the transitions not the "actual work".

the idea is just to have something print that we'll replace with the real implementation later to make sure the function is called when running.

```suggestion
	void suspendHardwareForDeepSleep() final
	{
		log_info("TO IMPLEMENT - configuring hardware for deepsleep");
	}
```

libs/RobotKit/include/RobotController.h
```diff
@@ -292,6 +305,8 @@ class RobotController : public interface::RobotController
 		stopActuators();
 	}
 
+	void suspendHardwareForDeepSleep() final { log_info("Entering deep sleep"); }
```
ladislas, line 308: to answer your previous comment and to make it more obvious

```suggestion
	void suspendHardwareForDeepSleep() final { 	log_info("TO IMPLEMENT - configuring hardware for deep sleep"); }

```

ladislas (approved):
LGTM 👍 - great work! 💪 🙏

PR #1241, opened 2023-01-04
:construction: (IMUKit): Update Mahony's sampling delay

HPezz:


ladislas:
experiment not successful

PR #1252, opened 2023-01-13
✨ (eventqueue): Add cancelLastCall

YannLocatelli:


drivers/CoreEventQueue/include/CoreEventQueue.h
```diff
@@ -18,19 +18,23 @@ class CoreEventQueue : public interface::EventQueue
...
-	void call(auto f, auto... params) { _event_queue.call(f, params...); }
+	void call(auto f, auto... params) { _event_id = _event_queue.call(f, params...); }
 
 	void call_every(std::chrono::duration<int, std::milli> duration, auto f, auto... params)
 	{
-		_event_queue.call_every(duration, f, params...);
+		_event_id = _event_queue.call_every(duration, f, params...);
 	}
 
+	void cancelLastCall() { _event_queue.cancel(_event_id); }
```
ladislas, line 28: I would just use `cancel`

```suggestion
	void cancel() { _event_queue.cancel(_event_id); }
```

in the future we could have `call` return the `id` to be able to `cancel(id)` and share an event queue between different callbacks

what happens if we call `call_every` multiple times? will the queue call all the callback at the desired interval? 

if so, then we should first `cancel` before using `_event_queue.call` or `_event_queue.call_every`

`_event_id` can be an optional to handle the case where cancel is called before `call_every` has been called.

drivers/CoreEventQueue/include/CoreEventQueue.h
```diff
@@ -18,19 +18,23 @@ class CoreEventQueue : public interface::EventQueue
 
 	void dispatch_forever() final;
 
-	void call(auto f, auto... params) { _event_queue.call(f, params...); }
+	void call(auto f, auto... params) { _event_id = _event_queue.call(f, params...); }
```
ladislas, line 21: here, the `_event_id` is not really needed and could even introduce bugs.

`_event_queue.call` will dispatch the event as soon as it's called .

per mbed documentation, when the event is dispatched, the id is not valid anymore and

> It is not safe to call cancel after an event has already been dispatched.
> id must be valid i.e. event must have not finished executing.

https://os.mbed.com/docs/mbed-os/v6.15/mbed-os-api-doxy/classevents_1_1_event_queue.html#a9ccb4f1222149f37161430fa2f1974e4

so calling `call` and then `cancel` might break the program.

cancel is only really needed for `call_every` or `call_in`

ladislas (changes requested):
LGTM but after reviewing mbed's documentation, I'm afraid that with the current implementation we might introduce bugs

ladislas (approved):
LGTM 👍

PR #1253, opened 2023-01-13
✅ Fix uninteresting mock function call

YannLocatelli:
2 drivers still have unintersting mock function call WARNING

- [CoreMotor](https://github.com/leka/LekaOS/tree/develop/drivers/CoreMotor) - `suspend` is called in constructor
- [CoreTouchSensor](https://github.com/leka/LekaOS/tree/develop/drivers/CoreTouchSensor) - `setAsInput` from DigitalIn and `setAsOutput` from DigitalOut are called in constructor


ladislas (approved):
LGTM 👍

PR #1255, opened 2023-01-13
♻️ Use std::function instead of mbed::Callback

YannLocatelli:
# Validations

- [x] BLE
- [x] Battery - onChargeDidStart
- [x] Battery - onChargeDidStop
- [x] RFID

### DIFF
--- a/drivers/CoreBattery/source/CoreBattery.cpp
+++ b/drivers/CoreBattery/source/CoreBattery.cpp
@@ -4,18 +4,22 @@
 
 #include "CoreBattery.h"
 
+#include "platform/Callback.h"
+
 #include "MathUtils.h"
 
 namespace leka {
 
-void CoreBattery::onChargeDidStart(mbed::Callback<void()> const &callback)
+void CoreBattery::onChargeDidStart(std::function<void()> const &callback)
 {
-	_charge_status_input.rise(callback);
+	_on_charge_did_start = callback;
+	_charge_status_input.rise(mbed::Callback<void()> {[this] { _on_charge_did_start(); }});
 }
 
-void CoreBattery::onChargeDidStop(mbed::Callback<void()> const &callback)
+void CoreBattery::onChargeDidStop(std::function<void()> const &callback)
 {
-	_charge_status_input.fall(callback);
+	_on_charge_did_stop = callback;
+	_charge_status_input.fall(mbed::Callback<void()> {[this] { _on_charge_did_stop(); }});
 }
 
 auto CoreBattery::voltage() -> float
--- a/drivers/CoreBufferedSerial/source/CoreBufferedSerial.cpp
+++ b/drivers/CoreBufferedSerial/source/CoreBufferedSerial.cpp
@@ -33,9 +33,10 @@ void CoreBufferedSerial::disable_input()
 	_serial.enable_input(false);
 }
 
-void CoreBufferedSerial::sigio(mbed::Callback<void()> func)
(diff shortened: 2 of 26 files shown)


drivers/CoreBattery/source/CoreBattery.cpp
```diff
@@ -4,18 +4,22 @@
...
 #include "MathUtils.h"
 
 namespace leka {
 
-void CoreBattery::onChargeDidStart(mbed::Callback<void()> const &callback)
+void CoreBattery::onChargeDidStart(std::function<void()> const &callback)
 {
-	_charge_status_input.rise(callback);
+	_on_charge_did_start = callback;
+	_charge_status_input.rise(mbed::Callback<void()> {[this] { _on_charge_did_start(); }});
```
ladislas, line 16: is it not possible to do just that? Do we really need to go from std::function to lambda to mbed::callback?

```suggestion
	_charge_status_input.rise([this] { _on_charge_did_start(); });
```

drivers/CoreBattery/source/CoreBattery.cpp
```diff
@@ -4,18 +4,22 @@
...
 }
 
-void CoreBattery::onChargeDidStop(mbed::Callback<void()> const &callback)
+void CoreBattery::onChargeDidStop(std::function<void()> const &callback)
 {
-	_charge_status_input.fall(callback);
+	_on_charge_did_stop = callback;
+	_charge_status_input.fall(mbed::Callback<void()> {[this] { _on_charge_did_stop(); }});
```
ladislas, line 22: same as above if possible

```suggestion
	_charge_status_input.fall([this] { _on_charge_did_stop(); });
```

drivers/CoreBufferedSerial/source/CoreBufferedSerial.cpp
```diff
@@ -33,9 +33,10 @@ void CoreBufferedSerial::disable_input()
 	_serial.enable_input(false);
 }
 
-void CoreBufferedSerial::sigio(mbed::Callback<void()> func)
+void CoreBufferedSerial::sigio(std::function<void()> const &callback)
 {
-	_serial.sigio(func);
+	_sigio_callback = callback;
+	_serial.sigio(mbed::Callback<void()> {[this] { _sigio_callback(); }});
```
ladislas, line 39: as we do in `CoreRFIDReaderCR95HF` mbed::callback could be removed

```suggestion
	_serial.sigio([this] { _sigio_callback(); });
```

libs/BLEKit/source/BLEKit.cpp
```diff
@@ -31,7 +33,7 @@ void BLEKit::init()
 
 void BLEKit::processEvents(BLE::OnEventsToProcessCallbackContext *context)
 {
-	_event_queue.callMbedCallback(mbed::callback(&context->ble, &BLE::processEvents));
+	_event_queue.call(mbed::Callback {&context->ble, &BLE::processEvents});
```
ladislas, line 36: also here, is mbed::Callback really necessary compared to a lambda?

ladislas (changes requested):
thanks for this PR! good to see mbed::callback going away in a lot of places :)

I've added a few suggestions where I think that mbed::callback can be completely removed and replaced by simple lambdas.

HPezz (commented):
Nice 👍 

- [x] Validated on robot

ladislas (approved):

PR #1258, opened 2023-01-16
✨ (config): Config with multiple values

YannLocatelli:
# Validation

- [x] Bootloader
- [x] Spike lk_config_kit

### DIFF
--- a/app/bootloader/main.cpp
+++ b/app/bootloader/main.cpp
@@ -158,10 +158,10 @@ namespace factory_reset {
 
 namespace config {
 
-	auto bootloader_version = Config {"/fs/sys/bootloader-version", bootloader::version};
+	auto bootloader_version = Config {"/fs/sys/bootloader-version", {bootloader::version}};
 	auto battery_hysteresis_offset =
-		Config {"/fs/etc/bootloader-battery_hysteresis", battery::default_hysteresis_offset};
-	auto factory_reset_limit = Config {"/fs/etc/bootloader-reboots_limit", factory_reset::default_limit};
+		Config {"/fs/etc/bootloader-battery_hysteresis", {battery::default_hysteresis_offset}};
+	auto factory_reset_limit = Config {"/fs/etc/bootloader-reboots_limit", {factory_reset::default_limit}};
 
 	auto configkit = ConfigKit {};
 
--- a/libs/ConfigKit/include/Config.h
+++ b/libs/ConfigKit/include/Config.h
@@ -4,25 +4,28 @@
 
 #pragma once
 
+#include <array>
+#include <cstdint>
 #include <filesystem>
 
 namespace leka {
 
+template <std::size_t SIZE = 1>
 struct Config {
   public:
-	explicit Config(const std::filesystem::path &path, uint8_t default_value = 0xFF)
-		: _path(_default_parent_path / path), _default_value(default_value)
+	explicit Config(const std::filesystem::path &path, std::array<uint8_t, SIZE> default_value = {})
+		: _path(_default_parent_path / path), kDefaultValue(default_value)
 	{
 		// nothing to do
 	}
 
 	[[nodiscard]] auto path() const -> std::filesystem::path { return _path; }
-	[[nodiscard]] auto default_value() const -> uint8_t { return _default_value; }
+	[[nodiscard]] auto default_value() const -> std::array<uint8_t, SIZE> { return kDefaultValue; }
 
   private:
 	const std::filesystem::path _default_parent_path = "/fs/etc";
 	const std::filesystem::path _path;
-	const uint8_t _default_value;
+	const std::array<uint8_t, SIZE> kDefaultValue {};
 };
 
 }	// namespace leka
--- a/libs/ConfigKit/include/ConfigKit.h
+++ b/libs/ConfigKit/include/ConfigKit.h
@@ -7,6 +7,7 @@
 #include <filesystem>
 
 #include "Config.h"
+#include "FileManagerKit.h"
 
 namespace leka {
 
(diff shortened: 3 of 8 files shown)


HPezz (approved):
Nice work

libs/ConfigKit/include/ConfigComplex.h
```diff
@@ -0,0 +1,34 @@
...
+#include <cstdint>
+#include <filesystem>
+#include <span>
+
+namespace leka {
+
+template <std::size_t SIZE>
+struct ConfigComplex {
+  public:
+	explicit ConfigComplex(const std::filesystem::path &path, std::span<uint8_t> default_value = {})
```
ladislas, line 17: here, as this is the ctor, it's strange to want to pass something that might be bigger or smaller than the size of the config.

If you create an instance, you know what you are doing, so you know the size. I don't see a reason to pass two values when the config is expecting 3 -- in this case, if the 3 value in not important, then the value itself should say that

with this, you can actually change the ctor to:

```suggestion
	explicit ConfigComplex(const std::filesystem::path &path, std::array<uint8_t, SIZE> default_value = {})
```

this will allow you to also remove the `min_size + std::copy` and just assign the values to the array like so (simpler example):

```cpp
template<std::size_t SIZE>
struct Config {
std::array<int, SIZE> arr;
};

auto conf = Config<3> {1, 2, 3};
```

https://gcc.godbolt.org/z/a3b3vePqe

libs/ConfigKit/include/ConfigComplex.h
```diff
@@ -0,0 +1,34 @@
...
+
+#include <array>
+#include <cstdint>
+#include <filesystem>
+#include <span>
+
+namespace leka {
+
+template <std::size_t SIZE>
+struct ConfigComplex {
```
ladislas, line 15: I'm wondering about the necessity to add yet another struct which would make Config vs ConfigComplex difficult to grasp.

In reality `Config == ConfigComplex<1>`

I would prefer we make `Config` evolve to be of arbitrary size and refactor it's current use with `Config<1>` rather than adding two different types of Config.

libs/ConfigKit/include/ConfigKit.h
```diff
@@ -7,15 +7,38 @@
...
+	{
+		if (FileManagerKit::File file {config.path(), "r"}; file.is_open()) {
+			auto data = std::array<uint8_t, N> {};
+			file.read(data);
+
+			return data;
+		}
+
+		return config.default_value();
+	}
```
ladislas, line 29: it's actually possible to have a function with different return types provided it's used in `constexpr` context

see https://gcc.godbolt.org/z/ren7Pa685

not sure if that would work there though, maybe for future improvements

libs/ConfigKit/include/Config.h
```diff
@@ -1,28 +1,31 @@
...
 
 	[[nodiscard]] auto path() const -> std::filesystem::path { return _path; }
-	[[nodiscard]] auto default_value() const -> uint8_t { return _default_value; }
+	[[nodiscard]] auto default_value() const -> std::array<uint8_t, SIZE> { return KDefaultValues; }
 
   private:
 	const std::filesystem::path _default_parent_path = "/fs/etc";
 	const std::filesystem::path _path;
-	const uint8_t _default_value;
+	const std::array<uint8_t, SIZE> KDefaultValues {};
```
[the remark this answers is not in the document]
ladislas: I don't know which is better : `KDefaultValues` or `KDefaultValue` with no `s`

we have `default_value()`

the question is: is the array the whole value? or do we store different values in the array?

libs/ConfigKit/include/ConfigKit.h
```diff
@@ -7,15 +7,38 @@
...
 
 namespace leka {
 
 class ConfigKit
 {
   public:
 	explicit ConfigKit() = default;
-	[[nodiscard]] auto read(Config const &config) const -> uint8_t;
-	[[nodiscard]] auto write(Config const &config, uint8_t data) const -> bool;
+	template <std::size_t N>
```
ladislas, line 18:
```suggestion
	template <std::size_t SIZE = 1>
```

libs/ConfigKit/include/Config.h
```diff
@@ -1,28 +1,31 @@
...
 
 #pragma once
 
+#include <array>
+#include <cstdint>
 #include <filesystem>
 
 namespace leka {
 
+template <std::size_t SIZE>
```
ladislas, line 13: we can make the default `1`

```suggestion
template <std::size_t SIZE = 1>
```

libs/ConfigKit/include/ConfigKit.h
```diff
@@ -7,15 +7,38 @@
...
+			auto data = std::array<uint8_t, N> {};
+			file.read(data);
+
+			return data;
+		}
+
+		return config.default_value();
+	}
+
+	template <std::size_t N>
```
ladislas, line 31:
```suggestion
	template <std::size_t N = 1> 
```

libs/ConfigKit/include/ConfigList.h
```diff
@@ -10,6 +10,7 @@
 
 namespace leka::config::bootloader {
 
-inline const auto battery_level_hysteresis = Config {"bootloader_battery_level_hysteresis", 42};
+inline const auto battery_level_hysteresis =
+	Config {"bootloader_battery_level_hysteresis", std::array<uint8_t, 1> {42}};
```
ladislas, line 14: same here to remove the need of std::array

app/bootloader/main.cpp
```diff
@@ -159,10 +159,11 @@ namespace factory_reset {
 
 namespace config {
 
-	auto bootloader_version = Config {"/fs/sys/bootloader-version", bootloader::version};
+	auto bootloader_version = Config {"/fs/sys/bootloader-version", std::array<uint8_t, 1> {bootloader::version}};
```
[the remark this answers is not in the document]
ladislas: just tested, this works:

```
auto bootloader_version = Config<1> {"/fs/sys/bootloader-version", {bootloader::version}};
```

and even 

```
auto bootloader_version = Config {"/fs/sys/bootloader-version", {bootloader::version}};
```

with the default size of `1` for `Config`

ladislas (changes requested):
great job merging both! 👍 

a few suggestions to make the code even easier to read :)

libs/ConfigKit/include/Config.h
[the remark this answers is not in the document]
YannLocatelli: Well, currently a config store a sole value and the next config is the name. The name is the value and each character are not relevant as values.

But another case is to consider to store a version as config, therefore the config can have 4 sub-config: major, minor, revision and build. However, this data should be cast in a structure and the array solely can't be considered as a value.

I suggest to keep it singular for now

libs/ConfigKit/include/ConfigKit.h
YannLocatelli: I tried something for read

libs/ConfigKit/include/Config.h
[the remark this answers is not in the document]
ladislas: I agree 👍

ladislas (approved):
LGTM 👍 great work!

ladislas (approved):

PR #1259, opened 2023-01-16
✨ (Firmware) - Add Format Factory

YannLocatelli:
# Validation

- [x] LekaOS-factory.bin
- [x] LekaOS-1.0.0.bin


HPezz (approved):
👍

libs/FirmwareKit/include/FirmwareKit.h
```diff
@@ -32,11 +34,12 @@ class FirmwareKit : public interface::FirmwareUpdate
 
 	auto isVersionAvailable(const Version &version) -> bool final;
 	auto loadUpdate(const Version &version) -> bool final;
+	auto loadFactory() -> bool final;
```
ladislas, line 37:
```suggestion
	auto loadFactoryFirmware() -> bool final;
```

libs/FirmwareKit/source/FirmwareKit.cpp
```diff
@@ -46,10 +46,18 @@ auto FirmwareKit::loadUpdate(const Version &version) -> bool
...
+	return load(path);
 }
 
-auto FirmwareKit::loadUpdate(const std::filesystem::path &path) -> bool
+auto FirmwareKit::loadFactory() -> bool
+{
+	if (auto factory_loaded = load(_config.factory_path); factory_loaded) {
+		return true;
+	}
+	return loadUpdate(Version {.major = 1, .minor = 0, .revision = 0});
```
ladislas, line 57: I'm not sure I understand what is happening here.

if you can load the factory image, you return true meaning that it's the one who is going to be installed, right?

but if you can't, why would you load an update instead?

ladislas (changes requested):
cool new feature as well, but we'll need to discuss this in our daily as I'm not sure I can grasp all the tenants et aboutissants of this change :)

libs/FirmwareKit/source/FirmwareKit.cpp
```diff
@@ -42,14 +42,23 @@ auto FirmwareKit::isVersionAvailable(const Version &version) -> bool
...
 	auto path = getPathOfVersion(version);
 
-	return loadUpdate(path);
+	return load(path);
 }
 
-auto FirmwareKit::loadUpdate(const std::filesystem::path &path) -> bool
+auto FirmwareKit::loadFactoryFirmware() -> bool
+{
+	if (auto does_factory_file_exist = _file.open(_config.factory_path); does_factory_file_exist) {
```
ladislas, line 54:
```suggestion
	if (auto factory_firmware_exists = _file.open(_config.factory_path); factory_firmware_exists) {
```

libs/FirmwareKit/source/FirmwareKit.cpp
```diff
@@ -42,14 +42,23 @@ auto FirmwareKit::isVersionAvailable(const Version &version) -> bool
...
+		return load(_config.factory_path);
+	}
+	return loadFirmware(
+		{.major = 1, .minor = 0, .revision = 0});	// ? Backward compatibility, factory file does not exist
```
ladislas, line 58:
```suggestion
	// ! IMPORTANT: BACKWARD COMPATIBILITY
	// ? This path is kept to handle the case where the bootloader has 
	// ? been updated but not the SD card, in which case the factory 
	// ? firmware might not exist. If it's the case, we must load the 
	// ? previous factory firmware which was LekaOS-1.0.0.bin
	// ? This should only happen when a robot is sent back to be fixed 
	// ? and used internally, but never for products in the field as 
	// ? the bootloader cannot change.
	return loadFirmware(
		{.major = 1, .minor = 0, .revision = 0});	
```

ladislas (approved):
LGTM 👍 just two small suggestions

ladislas (approved):
LGTM 👍

PR #1262, opened 2023-01-19
hugo/feature/Create CoreInterruptIn

HPezz:
- :building_construction: (Interface): Add InterruptIn interface
- :sparkles: (CoreInterruptIn): Create leka CoreInterruptIn wrapper
- :white_check_mark: (CoreInterruptIn): Add CoreInterruptIn_test


drivers/CoreInterruptIn/source/CoreInterruptIn.cpp
```diff
@@ -0,0 +1,24 @@
...
+{
+	_on_rise_callback = callback;
+	_interrupt_in.rise(mbed::Callback<void()> {[this] { _on_rise_callback(); }});
+}
+
+void CoreInterruptIn::fall(std::function<void()> const &callback)
+{
+	_on_fall_callback = callback;
+	_interrupt_in.fall(mbed::Callback<void()> {[this] { _on_fall_callback(); }});
+}
```
YannLocatelli, line 24: Regarding recent discovery, maybe should call the callback in event queue

```suggestion
void CoreInterruptIn::rise(std::function<void()> const &callback)
{
	_on_rise_callback = callback;
	_interrupt_in.rise(mbed::Callback<void()> {[this] { _event_queue.call( [this] { _on_rise_callback(); } ) }});
}

void CoreInterruptIn::fall(std::function<void()> const &callback)
{
	_on_fall_callback = callback;
	_interrupt_in.fall(mbed::Callback<void()> {[this] { _event_queue.call( [this] { _on_fall_callback(); } ) }});
}
```

drivers/CoreInterruptIn/include/CoreInterruptIn.h
```diff
@@ -0,0 +1,28 @@
...
+#pragma once
+
+#include "interface/drivers/InterruptIn.h"
+
+namespace leka {
+
+class CoreInterruptIn : public interface::InterruptIn
+{
+  public:
+	explicit CoreInterruptIn(PinName pin) : _pin(pin) {};
```
YannLocatelli, line 14: Does this work?

```suggestion
	explicit CoreInterruptIn(PinName pin) : _interrupt_in(pin) {};
```

drivers/CoreInterruptIn/include/CoreInterruptIn.h
```diff
@@ -0,0 +1,29 @@
...
+	explicit CoreInterruptIn(PinName pin) : _interrupt_in(pin) { _event_queue.dispatch_forever(); };
+
+	auto read() -> int final;
+
+	void rise(std::function<void()> const &callback) final;
+	void fall(std::function<void()> const &callback) final;
+
+  private:
+	CoreEventQueue _event_queue;
+	mbed::InterruptIn _interrupt_in;
```
ladislas, line 24: as there is only one pin:

```suggestion
	explicit CoreInterruptIn(PinName pin) : _pin(pin) { _event_queue.dispatch_forever(); };

	auto read() -> int final;

	void rise(std::function<void()> const &callback) final;
	void fall(std::function<void()> const &callback) final;

  private:
	CoreEventQueue _event_queue;
	mbed::InterruptIn _pin;
```

drivers/CoreInterruptIn/include/CoreInterruptIn.h
```diff
@@ -0,0 +1,29 @@
...
+  public:
+	explicit CoreInterruptIn(PinName pin) : _interrupt_in(pin) { _event_queue.dispatch_forever(); };
+
+	auto read() -> int final;
+
+	void rise(std::function<void()> const &callback) final;
+	void fall(std::function<void()> const &callback) final;
+
+  private:
+	CoreEventQueue _event_queue;
```
ladislas, line 23: as explained on slack, this should be removed.

interrupts are very low level hardware primitive used to react **instantly** when they trigger.

event queues are used to differ execution of slow parts of code outside of ISR context.

the fact that we need to use an event queue after an interrupt is an implementation detail and should be handled by the developper.

CoreInterruptIn must be kept as low level and fast as possible.

include/interface/drivers/InterruptIn.h
```diff
@@ -0,0 +1,27 @@
...
+
+class InterruptIn
+{
+  public:
+	virtual ~InterruptIn() = default;
+
+	virtual auto read() -> int = 0;
+
+	virtual void rise(std::function<void()> const &callback) = 0;
+	virtual void fall(std::function<void()> const &callback) = 0;
```
ladislas, line 24: i would add an `on`

```suggestion
	virtual void onRise(std::function<void()> const &callback) = 0;
	virtual void onFall(std::function<void()> const &callback) = 0;
```

ladislas (changes requested):
LGTM 👍 
except for the CoreEventQueue which must be removed

ladislas (approved):
LGTM 👍

PR #1264, opened 2023-01-20
⚡️ (video): Default start of path for displayed assets

YannLocatelli:
# Validation

- [x] LekaOS (RobotController)
- [x] Spike lk_lcd

### DIFF
--- a/libs/ActivityKit/source/ActivityKit.cpp
+++ b/libs/ActivityKit/source/ActivityKit.cpp
@@ -37,9 +37,9 @@ void ActivityKit::stop()
 void ActivityKit::displayMainMenu(const MagicCard &card)
 {
 	if (card.getLanguage() == MagicCard::Language::fr_FR) {
-		_videokit.displayImage("/fs/home/img/system/robot-misc-choose_activity-fr_FR.jpg");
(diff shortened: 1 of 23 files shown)


ladislas (commented):
Thanks for the PR @YannLocatelli!

After reviewing, I'm wondering if it's the right way to handle the assets:

- everything is static, I'm surprised that "refactoring" the root of the path adds so many bytes
- I'm also worried that doing the `root / path` things now add a lot of runtime / dynamic memory allocation. There's no way that this is static as, because of the changes, the compiler doesn't know in advance the values
- `std::filesystem::path` is a class and there is no information about `constexpr` -- to solve the previous issue, we would need to make all of our code `constexpr`, which I'm not sure is possible
- even if we remove the root of the path, it leaves us with the rest to type, which means that we can still make typos and the developper has to know what is available and where --> there's not progressive discovery as you type

We had the same issue for our swift app -- in Swift the recommended way is to generate `structs` to act as namespaces to hold the string values.

https://github.com/SwiftGen/SwiftGen does that automatically for swift.

In C++ you would end up typing "fs::home::img::system::robot_smiling" and this would return the real path.

The namespace generation can be automated with a simple python script.

I had started working on one a few months ago, I'll review it and share is with you.

YannLocatelli:
See #1266

PR #1269, opened 2023-01-24
:sparkles: (makefile): Build firmware + os in parallel

ladislas:


YannLocatelli (approved):
Everything is fine to me 👍

# Validations

- [x] From Github actions
- [x] From local `make firmware`

PR #1271, opened 2023-01-25
♻️ (video): Forward declare CGColor and CGPixel struct

YannLocatelli:
# Validation

- [x] lk_lcd
- [x] LekaOS (clearScreen)

### DIFF
--- a/drivers/CoreVideo/include/interface/Font.hpp
+++ b/drivers/CoreVideo/include/interface/Font.hpp
@@ -4,27 +4,25 @@
 
 #pragma once
 
-// #include <cstdint>
+#include <cstdint>
 
-#include "CGColor.hpp"
-#include "CGPixel.hpp"
+namespace leka {
+
+struct CGColor;
+struct Character;
+
+}	// namespace leka
 
 namespace leka::interface {
 
 class Font
 {
   public:
-	struct Character {
-		CGPoint origin {};	 // Top left corner by convention
-		uint8_t ascii {};	 // From 0x20 to 0x7F
-	};
-
 	virtual ~Font() = default;
 
-	virtual void drawChar(Character character, CGColor foreground = CGColor::black,
-						  CGColor background = CGColor::white) = 0;
-	virtual void display(const char *text, uint32_t size, uint32_t starting_line, CGColor foreground = CGColor::black,
-						 CGColor background = CGColor::white)  = 0;
+	virtual void drawChar(Character character, CGColor foreground, CGColor background) = 0;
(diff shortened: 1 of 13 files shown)


drivers/CoreVideo/include/interface/Font.hpp
```diff
@@ -4,27 +4,25 @@
...
 
-// #include <cstdint>
+#include <cstdint>
 
-#include "CGColor.hpp"
-#include "CGPixel.hpp"
+namespace leka {
+
+struct CGColor;
+struct Character;
```
ladislas, line 12: Nice! 👍 

In the future, I still have the wish to put all those "user types" in the main `include` directory and have our code depend on it directly.

This is something discussed by Tony in the Cpp.chat episode https://www.youtube.com/watch?v=WRMhO9Gn5GM

Very interesting talk btw.

drivers/CoreVideo/include/interface/Font.hpp
```diff
@@ -4,27 +4,25 @@
...
-		uint8_t ascii {};	 // From 0x20 to 0x7F
-	};
-
 	virtual ~Font() = default;
 
-	virtual void drawChar(Character character, CGColor foreground = CGColor::black,
-						  CGColor background = CGColor::white) = 0;
-	virtual void display(const char *text, uint32_t size, uint32_t starting_line, CGColor foreground = CGColor::black,
-						 CGColor background = CGColor::white)  = 0;
+	virtual void drawChar(Character character, CGColor foreground, CGColor background) = 0;
```
ladislas, line 23: the same question about background/foreground can be asked here: could they be stored directly inside the struct?

drivers/CoreVideo/tests/CoreFont_test.cpp
```diff
@@ -143,7 +144,7 @@ TEST_F(CoreFontTest, drawCharacterWithColor)
 {
 	// ENHANCEMENT: Set pixels_lit by checking number of bit set.
 
-	CoreFont::Character character;
+	Character character {};
 	character.ascii			 = '.';
 	CGColor foreground_color = CGColor::pure_red;
 	CGColor background_color = CGColor::black;
```
ladislas, line 150: building on the previous comment, storing the colors inside the struct would allow us to do:

```suggestion
	Character character {.ascii = '.',
						 .foreground = CGColor::pure_red,
						 .background= CGColor::black,
						};
```

ladislas (approved):
LGTM 👍 

A few questions/suggestions but not needed for this PR. If it makes sense we can do this in a future PR.

Issue #924, opened 2022-06-28
SM & App - Add a "Go to Sleep" button


HPezz:
RECOMMENDED BY EXPERTS : Add a way to trigger sleep mode, making easier the end of an activity w/ Leka for the child.

YannLocatelli:
Related to

- #1283

YannLocatelli:
Discussed during Educational Development Retrospective of 23/02/2023 this topic is not a necessity for Educational Team.

PR #1283, opened 2023-02-02
✨ (rc): Add Go To Sleep

YannLocatelli:
# Related

- #924

### DIFF
--- a/libs/BLEKit/include/BLEServiceMonitoring.h
+++ b/libs/BLEKit/include/BLEServiceMonitoring.h
@@ -29,13 +29,8 @@ class BLEServiceMonitoring : public interface::BLEService
 		sendData(data);
 	}
 
-	auto isScreensaverEnable() const -> bool { return screensaver_enable; }
-
 	void onDataReceived(const data_received_handle_t &params) final
 	{
-		if (params.handle == screensaver_enable_characteristic.getValueHandle()) {
-			screensaver_enable = static_cast<bool>(params.data[0]);
-		}
 		if (params.handle == soft_reboot_characteristic.getValueHandle()) {
 			soft_reboot = static_cast<bool>(params.data[0]);
 			if (soft_reboot && _on_soft_reboot) {
@@ -63,10 +58,6 @@ class BLEServiceMonitoring : public interface::BLEService
 		service::monitoring::characteristic::charging_status, &charging_status,
 		GattCharacteristic::BLE_GATT_CHAR_PROPERTIES_NOTIFY};
 
-	bool screensaver_enable {true};
-	WriteOnlyGattCharacteristic<bool> screensaver_enable_characteristic {
-		service::monitoring::characteristic::screensaver_enable, &screensaver_enable};
-
 	bool soft_reboot {false};
 	WriteOnlyGattCharacteristic<bool> soft_reboot_characteristic {service::monitoring::characteristic::soft_reboot,
 																  &soft_reboot};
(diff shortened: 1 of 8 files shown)


libs/BLEKit/include/BLEServiceMonitoring.h
```diff
@@ -52,6 +61,8 @@ class BLEServiceMonitoring : public interface::BLEService
 
 	void onSoftReboot(const std::function<void()> &callback) { _on_soft_reboot = callback; }
 
+	void registerGoToSleepRequested(const std::function<void()> &callback) { _on_go_to_sleep_callback = callback; }
```
ladislas, line 64: in English "register go to sleep requested" doesn't mean much.

I would say "register 'on go to sleep' callback"

question is: is the callback called:

- every time the app requested the robot to go to sleep?
- every time the robot actually goes to sleep?

is it possible for the request to no succeed? meaning that the robot will not go to sleep?

in this case, is the callback called?

to draw a parallel with iOS, a lot of Apple API have a notion of "tense":

- before action --> "view WILL appear" - `viewWillAppear`
- after action --> "view DID load" - `viewDidLoad`

it's important to know, from the API and the naming, if and when the registered callback will be called.

Taking this into account we could have:

```suggestion
	void registerWillGoToSleepCallback(const std::function<void()> &callback) { _on_go_to_sleep_callback = callback; }
```

```suggestion
	void registerDidGoToSleepCallback(const std::function<void()> &callback) { _on_go_to_sleep_callback = callback; }
```

libs/BLEKit/include/internal/ServicesCharacteristics.h
```diff
@@ -37,10 +37,10 @@ namespace monitoring {
...
 
 	namespace characteristic {
-		inline constexpr uint16_t charging_status	 = 0x6783;
-		inline constexpr uint16_t screensaver_enable = 0x8369;
-		inline constexpr uint16_t soft_reboot		 = 0x8382;
-		inline constexpr uint16_t hard_reboot		 = 0x7282;
+		inline constexpr uint16_t charging_status		   = 0x6783;
+		inline constexpr UUID::LongUUIDBytes_t go_to_sleep = "Go to Sleep";
+		inline constexpr uint16_t soft_reboot			   = 0x8382;
+		inline constexpr uint16_t hard_reboot			   = 0x7282;
```
ladislas, line 43: I suggest moving long UUIDs down and also using the upper case snake case `NAME_OF_SERVICE` to avoid confusion between lower/upper case and letter such as `i I l L` (i I l L)

```suggestion
		inline constexpr uint16_t charging_status		   = 0x6783;
		inline constexpr uint16_t soft_reboot			   = 0x8382;
		inline constexpr uint16_t hard_reboot			   = 0x7282;
		
		inline constexpr UUID::LongUUIDBytes_t go_to_sleep = "GO_TO_SLEEP";
```

libs/BLEKit/include/BLEServiceMonitoring.h
```diff
@@ -29,12 +29,21 @@ class BLEServiceMonitoring : public interface::BLEService
 		sendData(data);
 	}
 
-	auto isScreensaverEnable() const -> bool { return screensaver_enable; }
+	void resetGoToSleep()
```
ladislas, line 32: I'm not sure I understand what "reset go to sleep" actually mean in terms of what the method is doing.

libs/RobotKit/include/RobotController.h
```diff
@@ -482,6 +482,11 @@ class RobotController : public interface::RobotController
 		auto on_charge_did_stop = [this]() { raise(event::charge_did_stop {}); };
 		_battery.onChargeDidStop(on_charge_did_stop);
 
+		_service_monitoring.registerGoToSleepRequested([this] {
+			raise(event::go_to_sleep_requested {});
```
ladislas, line 486: If I understand correctly, this will trigger the sleeping animation with the LEDs and video. is that correct?

ladislas (changes requested):
Great work! 👍  I've made a few comments for suggestions and parts that need to be made a little clearer.

YannLocatelli:
Out of context and not requested by Educational Team.

Issue #822, opened 2022-06-02
BLEKit - Add isConnected


YannLocatelli:
BLEKit should be able to informs if BLE module is connected or not to a device

YannLocatelli:
Partially done with (BLEKit side)

- #1219

PR #1286, opened 2023-02-06
♻️ (rc): Use blekit BLEKit::isConnected instead of SM

YannLocatelli:
# Validation

- [x] LekaOS

# Related

- #822


YannLocatelli:
The blocking occurs at the unit test level, and the BLE (Bluetooth Low Energy) is poorly mocked at the events handlers level. To work around the issue, dependency injection is possible, but it would only be useful for unit tests and would not serve the functional purpose.

PR #1287, opened 2023-02-06
ladislas/feature/replace mahony with new fusion

ladislas:
- :truck: (spikes): Rename accel_gyro to sensors_imu_lsm6dsox
- :fire: (IMUKit): Remove Mahony implementation


spikes/lk_imu_kit/main.cpp
```diff
@@ -34,20 +34,22 @@ IMUKit imukit(imu::lsm6dsox);
...
 
+	imukit.stop();
 	imukit.init();
 	imukit.start();
 
+	const auto &[pitch, roll, yaw] = imukit.getEulerAngles();
+
 	while (true) {
-		auto [pitch, roll, yaw] = imukit.getAngles();
 		log_info("Pitch : %7.2f, Roll : %7.2f Yaw : %7.2f", pitch, roll, yaw);
```
ladislas, line 52: This is a funny thing. Sonarcloud says to use a const reference to avoid unnecessary copying of the values.

Doing so, we are able to call the function only once and still have the values updated in the while loop.

Not sure it's a good thing though: one might want to store the data and use the values later than ~20ms after the function call. In this case, the values will not hold the old ones, but new ones.

So Sonarcloud is not right there. It depends on the context. Maybe getEulerAngles should not return a reference but the struct to avoid the issue in any case.

ladislas (commented):
Personal comments

spikes/lk_imu_kit/main.cpp
HPezz: > one might want to store the data and use the values later than ~20ms after the function call. In this case, the values will not hold the old ones, but new ones
> Maybe getEulerAngles should not return a reference but the struct to avoid the issue in any case.

I don't see current application of a such behavior but this being said, and according to Motion Kit function's `run()` current implementation (getEulerAngles in a while loop), it might be better to return the struct.

HPezz (commented):
Nice job

One remark on the spike's start : yaw stays at 0.0 for several seconds then begin to really log value. 

Screenshot 2023-02-08 at 16 27 50

spikes/lk_imu_kit/main.cpp
ladislas: This has been implemented with no issues and is much less error prone

HPezz (commented):
LTGM 

Still have a question about testing. IMUKit unit tests are very light and functional tests are not updated yet. Is fusion inertial calibration which prevent to write generic functional tests or you plan to do some ?

libs/IMUKit/source/IMUKit.cpp
```diff
@@ -26,18 +70,45 @@ void IMUKit::stop()
...
-void IMUKit::computeAngles(const interface::LSM6DSOX::SensorData &imu_data)
+void IMUKit::drdy_callback(const interface::LSM6DSOX::SensorData data)
 {
-	auto xl_data			= std::make_tuple(imu_data.xl.x, imu_data.xl.y, imu_data.xl.z);
-	auto gy_data			= std::make_tuple(imu_data.gy.x, imu_data.gy.y, imu_data.gy.z);
-	constexpr auto mag_data = std::make_tuple(0.0F, 0.0F, 0.0F);
-	_mahony.update(xl_data, gy_data, mag_data);
-}
+	// ? Note: For a detailed explanation on the code below, checkout
+	// ? https://github.com/leka/LekaOS/tree/develop/spikes/lk_sensors_imu_lsm6dsox_fusion_calibration
```
YannLocatelli, line 87: I like this comment to keep trace of "original" or "inspiration" code, I suggest to also add a tag to the lib under the fusion's include as you added for the spike.

https://github.com/leka/LekaOS/blob/41f3e6c652fc5740d3dfe61fc7b92270bd6612f9/spikes/lk_sensors_imu_lsm6dsox_fusion_calibration/main.cpp#L14-L16

YannLocatelli (changes requested):
Validate with

- [x] lk_imu_kit spike

The whole integration looks good to me. I have some suggestions:

- Complete a unit test
- Fix failing test in CI for clang-format
- Add some "documentation"

Also found a potentiel bug

libs/IMUKit/external/CMakeLists.txt
```diff
@@ -1,10 +1,24 @@
...
+	PUBLIC
+	.
+)
 
-add_library(Fusion ${files})
+target_sources(Fusion
+	PRIVATE
+	fusion/FusionAhrs.c
+	fusion/FusionCompass.c
+	fusion/FusionOffset.c
```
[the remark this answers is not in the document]
ladislas: I prefer not to touch the files, I'll update the clang-format ignore in another PR instead

tests/unit/mocks/mocks/leka/LSM6DSOX.h
```diff
@@ -13,8 +13,14 @@ class LSM6DSOX : public interface::LSM6DSOX
 {
   public:
 	MOCK_METHOD(void, init, (), (override));
-	MOCK_METHOD(void, registerOnGyDataReadyCallback, (std::function<void(const SensorData &)> const &), (override));
 	MOCK_METHOD(void, setPowerMode, (PowerMode), (override));
+
+	void registerOnGyDataReadyCallback(drdy_callback_t const &cb) override { drdy_callback = cb; }
+
+	void call_drdy_callback(const SensorData &data) { drdy_callback(data); }
```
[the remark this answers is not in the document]
ladislas: fixed in https://github.com/leka/LekaOS/pull/1291

YannLocatelli (approved):
That's good for me 👍

HPezz (approved):
👍

PR #1292, opened 2023-02-17
hugo/feature/Update MotionKit to Fusion

HPezz:
MotionKit refactor following the previous IMUKit refactor with the Fusion algorithm and an asynchronous behavior. 

- [x] Rotation validated on robot
- [x]  Stabilization validated on robot


libs/IMUKit/tests/IMUKit_test.cpp
```diff
@@ -66,6 +66,42 @@ TEST_F(IMUKitTest, setOrigin)
...
+	auto angles_updated = imukit.getEulerAngles();
+
+	EXPECT_NE(angles_initial.pitch, angles_updated.pitch);
+	EXPECT_NE(angles_initial.roll, angles_updated.roll);
+	EXPECT_NE(angles_initial.yaw, angles_updated.yaw);
+}
+
+TEST_F(IMUKitTest, onDataReadyEulerAngleCallbackSetNullPtr)
+{
+	imukit.onEulerAnglesRdy(nullptr);
```
ladislas, line 103:
```suggestion
	imukit.onEulerAnglesRdy({});
```

libs/MotionKit/source/MotionKit.cpp
```diff
@@ -58,56 +56,54 @@ void MotionKit::startStabilisation()
...
+	_imukit.onEulerAnglesRdy(on_euler_angles_rdy_callback);
 }
 
 // LCOV_EXCL_START - Dynamic behavior, involving motors and time.
 void MotionKit::run()
 {
-	auto last_yaw			= kReferenceAngle;
-	auto rotations_executed = 0;
+	auto must_rotate = [&] { return _rotate_x_turns_requested && _rotations_executed < _rotations_to_execute; };
+	auto motion_end	 = [&] { return !_stabilisation_requested && !_rotate_x_turns_requested && !_target_not_reached; };
```
ladislas, line 75: I'd keep the same structure `must_rotate` vs `must_stop`

```suggestion
	auto must_rotate = [&] { return _rotate_x_turns_requested && _rotations_executed < _rotations_to_execute; };
	auto must_stop	 = [&] { return !_stabilisation_requested && !_rotate_x_turns_requested && !_target_not_reached; };
```

libs/MotionKit/source/MotionKit.cpp
```diff
@@ -58,56 +56,54 @@ void MotionKit::startStabilisation()
...
 
-	auto must_rotate = [&] { return _rotate_x_turns_requested && rotations_executed != _rotations_to_execute; };
+		_imukit.onEulerAnglesRdy(nullptr);
+		return;
+	}
 
 	auto check_complete_rotations_executed = [&](auto current_yaw) {
-		if (std::abs(last_yaw - current_yaw) >= 300.F) {
-			++rotations_executed;
+		if (std::abs(_previous_euler_angles.yaw - current_yaw) >= 300.F) {
```
ladislas, line 86: are you sure `300.F` is still working?

yaw is now `-180..+180`

you might need to map the values on `0..359`

it was already the case before, but I still can't wrap my head around this way of calculating rotations.

especially now that we don't reset the origin, if you start close to `~350`, it will count rotations for just a few degrees.

The right way to do it would be to accumulate the rotation angle over time and check if you have reached `~359`

it's possible to count clockwise and counter clockwise as well

ladislas (changes requested):
first review, thanks for the refactor. 👍 

I still have a lot of questions and I'm not sure the system is error proof. 

MotionKit tests must not rely on LSM6DSOX but must use an IMUKit mock

libs/MotionKit/source/MotionKit.cpp
```diff
@@ -58,56 +56,62 @@ void MotionKit::startStabilisation()
...
 
-	_rotate_x_turns_requested = false;
-	_rotations_to_execute	  = 0;
+	auto counting_rotations = [&](auto current_yaw) {
+		auto absolute_yaw_difference = std::abs(_euler_angles_previous.yaw - current_yaw);
+		if (absolute_yaw_difference >= 300.F) {
+			_rotations_executed += 360.F - absolute_yaw_difference;
+		} else {
+			_rotations_executed += absolute_yaw_difference;
+		}
```
ladislas, line 94: from sonarcloud

```suggestion
		if (auto abs_yaw_delta = std::abs(_euler_angles_previous.yaw - current_yaw); abs_yaw_delta >= 300.F) {
			_rotations_executed += 360.F - abs_yaw_delta;
		} else {
			_rotations_executed += abs_yaw_delta;
		}
```

libs/MotionKit/source/MotionKit.cpp
```diff
@@ -58,56 +56,62 @@ void MotionKit::startStabilisation()
...
-		rtos::ThisThread::sleep_for(70ms);
+		_euler_angles_previous = _euler_angles_current;
+		return;
 	}
 
-	if (_on_rotation_ended_callback != nullptr) {
-		_on_rotation_ended_callback();
-	}
+	if (_stabilisation_requested || _target_not_reached) {
+		_euler_angles_current = _imukit.getEulerAngles();
```
ladislas, line 109: same, no need for that here

ladislas (changes requested):
Much better, thanks! 
Added a few suggestions.

We'll discuss tests tomorrow :)

libs/MotionKit/include/MotionKit.hpp
```diff
@@ -47,9 +48,12 @@ class MotionKit
...
 	bool _rotate_x_turns_requested = false;
 
-	EulerAngles _euler_angles_previous {};
+	interface::EulerAngles _euler_angles_previous
+	{
+	};
 
-	const float kPIDMaxValue				  = 1.8F;
+	// ? Calculated with following PID parameters : ((270-180)*Kp + (360-180)*Kd)/KDeltaT
+	const float kEntryLimitSpeed			  = 3.15F;
```
ladislas, line 56: I don't understand the name of the variable

`entry` is usually used for rooms like `entry hall` (hall d'entrée)

I think you mean `input` (vs output)

```suggestion
const float kInputSpeedLimit = 3.15F;
```

libs/MotionKit/include/PID.hpp
```diff
@@ -18,7 +18,7 @@ class PID
 	void setTargetYaw(float angle);
 	auto processPID([[maybe_unused]] float pitch, [[maybe_unused]] float roll, float yaw)
 		-> std::tuple<float, Rotation>;
-	auto processPIDError(float _error_position_current) -> float;
+	auto processPIDByError(float _error_position_current) -> float;
```
ladislas, line 21: what does `by error` mean? calculate? what's the output?

include/interface/libs/IMUKit.h
```diff
@@ -0,0 +1,32 @@
+// Leka - LekaOS
```
ladislas, line 1: make `.hpp`

include/interface/libs/IMUKit.h
```diff
@@ -0,0 +1,32 @@
...
+#pragma once
+
+#include <functional>
+namespace leka::interface {
+
+struct EulerAngles {
+	float pitch;
+	float roll;
+	float yaw;
+};
```
ladislas, line 14: can you move it out of `interface` so that we don't need to change code using them.

They could be forward declared here actually, to keep the definition in IMUKit.hpp

include/interface/libs/IMUKit.h
```diff
@@ -0,0 +1,32 @@
...
+
+class IMUKit
+{
+  public:
+	using angles_ready_callback_t = std::function<void(const EulerAngles &)>;
+
+	virtual ~IMUKit() = default;
+
+	virtual void init()	 = 0;
```
ladislas, line 23: not sure we need an `init()` -- better keep it to the minimal, MotionKit should not depend on `init()`

libs/IMUKit/include/IMUKit.hpp
```diff
@@ -5,32 +5,31 @@
...
+	[[nodiscard]] auto getEulerAngles() const -> interface::EulerAngles final;
 
   private:
 	void drdy_callback(interface::LSM6DSOX::SensorData data);
 
 	interface::LSM6DSOX &_lsm6dsox;
-	EulerAngles _euler_angles {};
+	interface::EulerAngles _euler_angles
+	{
+	};
```
ladislas, line 31: this formatting here seems like a bug from clang-format, I'll look into it in the future

libs/MotionKit/include/MotionKit.hpp
```diff
@@ -28,14 +29,14 @@ class MotionKit
 	void stop();
 
   private:
-	void iterateOverRotation(const EulerAngles &angles, Rotation direction = Rotation::clockwise);
+	void iterateOverRotation(const interface::EulerAngles &angles, Rotation direction = {});
```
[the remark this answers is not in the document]
ladislas:
```suggestion
	void iterateOverRotation(const interface::EulerAngles &angles, Rotation direction);
```

libs/MotionKit/include/PID.hpp
```diff
@@ -30,15 +32,15 @@ class PID
...
-	float _derivative			  = 0.F;
+	const float kDeltaT		 = 20.F;
+
+	float _target_yaw {};
+
+	float _error_position_total = 0.F;
+	float _error_position_last	= 0.F;
+	float _proportional			= 0.F;
+	float _integral				= 0.F;
+	float _derivative			= 0.F;
```
ladislas, line 43: could be made a struct as they are all related, at least the last 3

libs/MotionKit/source/MotionKit.cpp
```diff
@@ -58,61 +54,74 @@ void MotionKit::startStabilisation()
...
+	_euler_angles_previous = _imukit.getEulerAngles();
+	_pid.setTargetYaw(_euler_angles_previous.yaw);
 
 	_target_not_reached		  = false;
 	_stabilisation_requested  = true;
 	_rotate_x_turns_requested = false;
 
-	_event_loop.start();
+	auto on_euler_angles_rdy_callback = [this](const interface::EulerAngles &euler_angles) {
+		iterateOverRotation(euler_angles);
```
ladislas, line 65: it's quite complicated to understand why it is sometimes called with `direction` and sometimes without.

it would be better to have two different functions to make it clearer for the user/reader

libs/MotionKit/source/MotionKit.cpp
```diff
@@ -58,61 +54,74 @@ void MotionKit::startStabilisation()
...
-	_imukit.stop();
 }
 
 auto MotionKit::mapSpeed(float speed) const -> float
 {
-	return utils::math::map(speed, 0.F, kPIDMaxValue, kMinimalViableRobotPwm, kPwmMaxValue);
+	return utils::math::map(speed, 0.F, kEntryLimitSpeed, kMinimalViableRobotPwm, kPwmMaxValue);
 }
 
 void MotionKit::executeSpeed(float speed, Rotation direction)
```
ladislas, line 127: `executeSpeed` does not sound like real English

```suggestion
void MotionKit::setMotorsSpeedAndDirection(float speed, Rotation direction)
```

tests/unit/mocks/mocks/leka/IMUKit.h
```diff
@@ -0,0 +1,29 @@
+// Leka - LekaOS
```
ladislas, line 1: make `.hpp`

ladislas (approved):
thanks for the refactor 👍 

I think we're good for the time being.

I've made some suggestions/refactors for 1.5.0 to allow us to do what we talked about in our weekly #1308 and provide a lot of different controls

MotionKit won't need to depend on IMUKit and Motors, only the controls will depend on those

libs/MotionKit/source/RotationControl.cpp
```diff
@@ -0,0 +1,53 @@
...
+	return mapSpeed(speed);
+}
+
+auto RotationControl::calculateYawRotation(float previous_yaw, float yaw) -> float
+{
+	auto abs_yaw_delta = std::abs(previous_yaw - yaw);
+	if (abs_yaw_delta >= 300.F) {
+		return 360.F - abs_yaw_delta;
+	}
+	return abs_yaw_delta;
```
ladislas, line 43:
```suggestion
auto RotationControl::calculateYawRotation(float previous, float current) -> float
{
	auto abs_delta = std::abs(previous - current);
	if (abs_delta >= 300.F) {
		return 360.F - abs_delta;
	}
	return abs_delta;
```

libs/MotionKit/source/MotionKit.cpp
```diff
@@ -58,73 +52,58 @@ void MotionKit::startStabilisation()
...
 
-	_rotate_x_turns_requested = false;
-	_rotations_to_execute	  = 0;
+	_angle_rotation_sum += _rotation_control.calculateYawRotation(_euler_angles_previous.yaw, angles.yaw);
 
-	while (_stabilisation_requested || _target_not_reached) {
-		auto [pitch, roll, yaw] = _imukit.getEulerAngles();
-		auto [speed, rotation]	= _pid.processPID(pitch, roll, yaw);
+	if (_rotate_x_turns_requested && _target_not_reached) {
+		auto speed = _rotation_control.processRotationAngle(_angle_rotation_target, _angle_rotation_sum);
```
ladislas, line 82: why do you need this to be here? 

the RotationControl should handle this logic, not MotionKit

ladislas: the whole function should be in RotationControl

you push the raw yaw value as input and get speed + direction as output

(before we move all that to the control class completely)

libs/MotionKit/source/MotionKit.cpp
```diff
@@ -4,52 +4,46 @@
...
 	_timeout.start(10s);
 
-	_event_loop.start();
+	auto on_euler_angles_rdy_callback = [this, direction](const EulerAngles &euler_angles) {
+		processAngleForRotation(euler_angles, direction);
+	};
+
+	_imukit.onEulerAnglesReady(on_euler_angles_rdy_callback);
 
 	_on_rotation_ended_callback = on_rotation_ended_callback;
```
ladislas, line 48: All of this should be part of RotationControl

ladislas (changes requested):
Seems like a few things have changed between yesterday and now, I've made new change requests.

to me, too much "rotation" logic is still present in MotionKit whereas it should be moved to RotationControl.

HPezz:
PR split into: 

1.  #1314 , a refactor PR of MptionKit including Rotation control only 
2. #1315 , adding stabilization control & yaw stabilization method in MotionKit

ladislas:
replaced by #1314 and #1315

PR #1309, opened 2023-03-03
✨ (sm): Add transition from AutonomousActivity to Sleeping

YannLocatelli:
## Validation

- [x] LekaOS

### DIFF
--- a/libs/RobotKit/include/RobotController.h
+++ b/libs/RobotKit/include/RobotController.h
@@ -440,8 +440,7 @@ class RobotController : public interface::RobotController
 		}
 
 		if (NOT_is_playing && is_autonomous_mode) {
-			auto before_process_card_callback = [this] { resetAutonomousActivitiesTimeout(); };
-			_activitykit.start(card, before_process_card_callback);
+			_activitykit.start(card);
 		}
 	}
 
@@ -459,6 +458,8 @@ class RobotController : public interface::RobotController
 			_service_magic_card.setMagicCard(card);
 			onMagicCardAvailable(card);
 		});
+
+		_activitykit.registerBeforeProcessCallback([this] { resetAutonomousActivitiesTimeout(); });
 
 		_battery_kit.onDataUpdated([this](uint8_t level) {
 			auto is_charging = _battery.isCharging();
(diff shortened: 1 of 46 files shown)


libs/RobotKit/include/RobotController.h
```diff
@@ -425,6 +425,10 @@ class RobotController : public interface::RobotController
 		}
 
 		if (NOT_is_playing && is_autonomous_mode) {
+			// TODO: Determine behavior
+			// 1. Do nothing - Sleeping duration is for Main Menu + 1 Activity (current implementation)
+			// 2. Restart - Sleeping duration is for 1 Activity
+			// 3. Disable - No sleeping duration
```
HPezz, line 431: I would propose to restart timer on magic card available, without distinction, whether it is dice card, number or whatever the card, it restart the timer. If a child have a trouble with an activity, this last can be longer than 10 minutes. The only case of Leka being asleep should be when no card is given to Leka for 10 minutes

libs/RobotKit/include/StateMachine.h
```diff
@@ -101,7 +101,8 @@ namespace sm::action {
 	};
 
 	struct start_sleep_timeout {
-		auto operator()(irc &rc) const { rc.startSleepTimeout(); }
+		int duration = 60;
```
HPezz, line 104: Do we have to specify a value here ?

YannLocatelli: If not specified, the default value is 0, so instead of having an instant timeout there is a default one to 60s, just in case

libs/RobotKit/include/RobotController.h
ladislas: I agree with that

libs/RobotKit/include/RobotController.h
```diff
@@ -82,13 +82,13 @@ class RobotController : public interface::RobotController
 		rtos::ThisThread::sleep_for(3s);
 	}
 
-	void startSleepTimeout() final
+	void startSleepTimeout(int duration) final
```
ladislas, line 85: let use `<chrono>` here

```suggestion
	void startSleepTimeout(std::chrono::seconds duration) final
```

ladislas (approved):
LGTM 👍 

I agree with @HPezz regarding reseting the timeout on every card

+ small suggestion to use chrono instead of ints for duration

YannLocatelli:
closed in favor of

- #1322

Issue #1338, opened 2023-12-13
CI is broken


YannLocatelli:
2 kind of issues seems to break the CI

---

## import of pyelftools

- [x] Resolved

<details><summary>Logs</summary>

```
🏃 Running configuration script for target LEKA_V1_2_DEV 📝
python3 /home/runner/work/LekaOS/LekaOS/cmake/scripts/configure_cmake_for_target.py LEKA_V1_2_DEV -p /home/runner/work/LekaOS/LekaOS/_build/LEKA_V1_2_DEV/cmake_config -a /home/runner/work/LekaOS/LekaOS/config/mbed_app.json
Traceback (most recent call last):
  File "/home/runner/work/LekaOS/LekaOS/cmake/scripts/configure_cmake_for_target.py", line 21, in <module>
    from tools import build_api, options
  File "/home/runner/work/LekaOS/LekaOS/extern/mbed-os/tools/build_api.py", line 36, in <module>
    from .arm_pack_manager import Cache
  File "/home/runner/work/LekaOS/LekaOS/extern/mbed-os/tools/arm_pack_manager/__init__.py", line 29, in <module>
    from tools.flash_algo import PackFlashAlgo
  File "/home/runner/work/LekaOS/LekaOS/extern/mbed-os/tools/flash_algo/__init__.py", line 33, in <module>
    from elftools.common.py3compat import bytes2str
ModuleNotFoundError: No module named 'elftools.common.py3compat'
Failed to import the Mbed OS Python configuration system!
Be sure that you have installed all required python modules, see  /home/runner/work/LekaOS/LekaOS/extern/mbed-os /requirements.txt
make[1]: *** [Makefile:1[46](https://github.com/leka/LekaOS/actions/runs/7169779695/job/19520951046#step:6:47): config_cmake_target] Error 1
make[1]: Leaving directory '/home/runner/work/LekaOS/LekaOS'
make: *** [Makefile:140: config] Error 2
Error: Process completed with exit code 2.
```
From https://github.com/leka/LekaOS/actions/runs/7169779695/job/19520951046
</details>

Might be resolved with ~~#1339~~ see https://github.com/leka/mbed-os/commit/6dd070fd378035a1a90cccebb9300f2958a54f8b

---

## unit tests with clang

- [x] Resolved

Workflows using clang seem to fail

<details><summary>Logs</summary>

```
[...]

PR #1339, opened 2023-12-13
👷 (tools): Set mbed dependencies to specific versions

YannLocatelli:
Same as mbed-os requirements.txt

pyelftools 0.30 is responsible of CI failure

Related to
- #1338


YannLocatelli:
See #1342

PR #1356, opened 2024-01-17
Yann/feature/ble/allow larger packet reception

YannLocatelli:
## Requirements

- [x] (optional) #1353

### DIFF
--- a/config/mbed_app.json
+++ b/config/mbed_app.json
@@ -15,6 +15,7 @@
 			"bluenrg_ms.SPI_nCS": "BLE_SPI_NSS",
 			"bluenrg_ms.SPI_SCK": "BLE_SPI_SCK",
 			"cordio.desired-att-mtu": 251,
+			"cordio.max-prepared-writes": 1,
 			"sd.SPI_CS": "SD_SPI_CS",
 			"sd.SPI_MISO": "SD_SPI_MISO",
 			"sd.SPI_MOSI": "SD_SPI_MOSI",
--- a/config/mbed_app.json
+++ b/config/mbed_app.json
@@ -15,6 +15,7 @@
 			"bluenrg_ms.SPI_nCS": "BLE_SPI_NSS",
 			"bluenrg_ms.SPI_SCK": "BLE_SPI_SCK",
 			"cordio.desired-att-mtu": 251,
+			"cordio.rx-acl-buffer-size": 259,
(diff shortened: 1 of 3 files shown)


config/mbed_app.json
```diff
@@ -15,6 +15,7 @@
 			"bluenrg_ms.SPI_nCS": "BLE_SPI_NSS",
 			"bluenrg_ms.SPI_SCK": "BLE_SPI_SCK",
 			"cordio.desired-att-mtu": 251,
+			"cordio.max-prepared-writes": 1,
```
ladislas, line 19: @YannLocatelli can you explain why the default value of `4` was not working and why the value of `1` is better? 

my understanding is that a lower value might impact the efficiency of operations requiring multiple attribute writes.

is `1` used to avoid more than one thing happening at once? and to make sure each command are processed one at a time?

can you explain the issue you were encountering?

config/mbed_app.json
```diff
@@ -15,6 +15,7 @@
 			"bluenrg_ms.SPI_nCS": "BLE_SPI_NSS",
 			"bluenrg_ms.SPI_SCK": "BLE_SPI_SCK",
 			"cordio.desired-att-mtu": 251,
+			"cordio.rx-acl-buffer-size": 189,
```
ladislas, line 18: same here, it's not clear to me why the default value `70` was not working and why the value `189` is the one chosen.

my understanding (but I might be wrong) is that the value must be bigger than the `desired-att-mtu` size.

what would happen with a value of `256`?

ladislas (changes requested):
I need better understanding of the values chosen, why the default ones where not good, the issues encountered and how you came up with the new values.

config/mbed_app.json
YannLocatelli: Never said it does not work
It is a size for a "packet" and it was downed to 189 to fit with iPad max buffer size (modulo some header or wrapper related to BLE protocol probably)
As you suggest with ChatGPT, value of 259 works

config/mbed_app.json
YannLocatelli: I talked about it few months ago, with value of 4 it works but we have to handle it differently.

You receive 4 packets at the same time in the robot and since you write directly in the file the data you received, you have to handle writing + ordering at the same time
I tried some solutions but never succeed, if you have any idea, let me know.

